PR #8, opened 2016-08-01
switch left_error and right_error. Not sending empty messages

jim-v:
It seems to work well on a robot after these minor changes.


rohbotics:
Great

PR #14, opened 2016-11-27
Try to get firmware version, throw after 10 tries

### LINKED ISSUE #13: Exit with error if firmware version message not received
This would save time diagnosing a dead PSOC or broken serial connection.

rohbotics:
Fixes #13 

Do we really want to exit, or just log?


jim-v (approved):
It works as expected. 

Note that the initialization of int firmware version gives a compiler warning.

PR #20, opened 2017-03-05
Clean out serial loop

rohbotics:
Commands are now transmitted directly, without going through a queue.
The reading loop properly waits for serial data, and handles the input better.


rohbotics:
@jim-v can you take a look?

jim-v (approved):
Sorry I didn't notice it earlier.   It looks reasonable to me.

Issue #21, opened 2017-04-20
Don't print error messages at startup


jim-v:
We should suppress the first few of these messages to avoid confusing the user:

ERROR] [1492649264.180968214]: REJECT: ? ??
[ERROR] [1492649264.181731857]: REJECT: ??
[ERROR] [1492649264.182024565]: REJECT: ??
[ERROR] [1492649264.182270657]: REJECT: ?
[ERROR] [1492649264.182507271]: REJECT: 
[ERROR] [1492649264.183878308]: REJECT: ?
[ERROR] [1492649264.184854712]: REJECT: 
[ERROR] [1492649264.185267575]: REJECT: 
[ERROR] [1492649264.185616792]: REJECT: 
[ERROR] [1492649264.190286517]: REJECT: 
[ERROR] [1492649264.191024379]: REJECT: ?p?? ?~
[ERROR] [1492649264.191544742]: REJECT: p?? ?~
[ERROR] [1492649264.192052813]: REJECT: ?? ?~
[ERROR] [1492649264.192717186]: REJECT: ? ?~
[ERROR] [1492649264.193275153]: REJECT: ?~
[ERROR] [1492649264.193801349]: REJECT: ?~
[ERROR] [1492649264.194099213]: DESERIALIZATION ERROR! - 2
[ERROR] [1492649264.194826294]: REJECT: 
[ERROR] [1492649264.200572786]: REJECT: <0
[ERROR] [1492649264.201176013]: REJECT: 0
[ERROR] [1492649264.201583720]: REJECT: 
[ERROR] [1492649264.201962312]: REJECT: 
[ERROR] [1492649264.202482936]: REJECT: 
[ERROR] [1492649264.202893351]: REJECT: 
[ERROR] [1492649264.203345329]: REJECT: ?~<0
[ERROR] [1492649264.712537218]: Unsupported dynamic_reconfigure level 4294967295

PR #22, opened 2017-04-22
Supress some potentially confusing warnings

jim-v:
Address #21 by suppressing the first few serial warnings.


rohbotics (approved):
Sorry for the delay on the review, it slipped through the cracks

PR #27, opened 2017-06-24
Update README.md

davecrawley:
Hi guys I am trying to put current knowledge in the documentation. I have also put in what I think are appropriate headers so that we can start to fill out the sections that we need on this one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,34 @@
 # ubiquity_motor
 [![Build Status](https://travis-ci.org/UbiquityRobotics/ubiquity_motor.svg?branch=indigo-devel)](https://travis-ci.org/UbiquityRobotics/ubiquity_motor)
 
-Package that provides a ROS interface for the motors in UbiquityRobotics robots
+Introduction
+
+This is a Package that provides a ROS interface for the motors in UbiquityRobotics robots
+
+Installation - From Binary
+
+This package may be installed from binaries for both x86 and ARM architectures. To install the ppa for this package use the following command:
+
+[insert ppa command here]
+
+The package may be installed with:
+
+ sudo apt-get install ros-kinetic-ubiquity-motor
+
+Installation - From Source
+
+[to be completed]
+
+Running your ubiquity motor controller  on the native rasperry pi interface with ubiquity_motor
+
+[to be completed]
+
+Running your ubiquity motor controller over USB using an FTDI USB-serial device with ubiquity_motor
+
+Ubiquity motors motor controller has a connector that is designed to operate with FTDI USB-serial devices thus making it possible to add a USB to serial device on the robot and make it run over USB. In order to run ubiquity_motor with an FTDI it is necessary to specify the FTDI device e.g. /dev/ttyUSB0 in magni_bringup/param/base.yaml.
+
+Running your ubiquity motor controller using a PCI to serial connector (e.g. for NUC or other desktop computers)
+
+[to be completed]
+
+Known Issues


README.md
```diff
@@ -1,4 +1,34 @@
...
-Package that provides a ROS interface for the motors in UbiquityRobotics robots
+Introduction
+
+This is a Package that provides a ROS interface for the motors in UbiquityRobotics robots
+
+Installation - From Binary
+
+This package may be installed from binaries for both x86 and ARM architectures. To install the ppa for this package use the following command:
+
+[insert ppa command here]
```
jim-v, line 12: The PPA is not necessary; the package is in the main ROS repos.

rohbotics: Agree with @jim-v the ppa lines should be removed. They can be replaced by something like:

The packages are available from the OSRF build farm.  See the ROS installation instructions for more details http://wiki.ros.org/kinetic/Installation/Ubuntu.

README.md
```diff
@@ -1,4 +1,34 @@
...
+
+Running your ubiquity motor controller over USB using an FTDI USB-serial device with ubiquity_motor
+
+Ubiquity motors motor controller has a connector that is designed to operate with FTDI USB-serial devices thus making it possible to add a USB to serial device on the robot and make it run over USB. In order to run ubiquity_motor with an FTDI it is necessary to specify the FTDI device e.g. /dev/ttyUSB0 in magni_bringup/param/base.yaml.
+
+Running your ubiquity motor controller using a PCI to serial connector (e.g. for NUC or other desktop computers)
+
+[to be completed]
+
+Known Issues
```
rohbotics, line 34: We should just refer to the issue tracker here.

README.md
```diff
@@ -1,4 +1,34 @@
...
+
+[to be completed]
+
+Running your ubiquity motor controller over USB using an FTDI USB-serial device with ubiquity_motor
+
+Ubiquity motors motor controller has a connector that is designed to operate with FTDI USB-serial devices thus making it possible to add a USB to serial device on the robot and make it run over USB. In order to run ubiquity_motor with an FTDI it is necessary to specify the FTDI device e.g. /dev/ttyUSB0 in magni_bringup/param/base.yaml.
+
+Running your ubiquity motor controller using a PCI to serial connector (e.g. for NUC or other desktop computers)
+
+[to be completed]
```
rohbotics, line 32: I don't think that we need to have all these separate sections for running on every platform. All the configuration for the node is in magni_robot. If the robot that a customer is running is not a standard Magni, they are going to have to customize magni_robot anyway. 

In here we should simply have one line about how the configuration in magni_robot may need to be changed to support having a FTDI (/dev/ttyUSB0) or internal (/dev/ttyS0) serial port.

README.md
```diff
@@ -1,4 +1,34 @@
...
+
+Running your ubiquity motor controller  on the native rasperry pi interface with ubiquity_motor
+
+[to be completed]
+
+Running your ubiquity motor controller over USB using an FTDI USB-serial device with ubiquity_motor
+
+Ubiquity motors motor controller has a connector that is designed to operate with FTDI USB-serial devices thus making it possible to add a USB to serial device on the robot and make it run over USB. In order to run ubiquity_motor with an FTDI it is necessary to specify the FTDI device e.g. /dev/ttyUSB0 in magni_bringup/param/base.yaml.
```
rohbotics, line 28: Let's not refer to specific hardware in here, the idea is that this node will interface with the motor control system on both Magni and Loki.

Rohan

README.md
```diff
@@ -1,4 +1,34 @@
 # ubiquity_motor
 [![Build Status](https://travis-ci.org/UbiquityRobotics/ubiquity_motor.svg?branch=indigo-devel)](https://travis-ci.org/UbiquityRobotics/ubiquity_motor)
 
-Package that provides a ROS interface for the motors in UbiquityRobotics robots
+Introduction
+
+This is a Package that provides a ROS interface for the motors in UbiquityRobotics robots
```
rohbotics, line 6: Nit: Package should not be capitalized, and Ubiquity Robotics should be two words in text.

rohbotics (changes requested):

README.md
davecrawley: Agree can you just change the text the way you want it?

README.md
davecrawley: Agree can you just change the text the way you want it?

README.md
davecrawley: Agree can you just change the text the way you want it?

README.md
davecrawley: Agree can you just change the text the way you want it?

README.md
davecrawley: Agree can you just change the text the way you want it? Am I missing something here?

PR #29, opened 2017-08-24
Publish battery voltage messages

### LINKED ISSUE #12: Publish power status messages
Finish figuring out relationship between ADC readings and currents and put into the CPP

jim-v:
Closes #12

### DIFF
--- a/src/motor_hardware.cc
+++ b/src/motor_hardware.cc
@@ -153,11 +153,11 @@ void MotorHardware::readInputs() {
                     bstate.charge = std::numeric_limits<float>::quiet_NaN();
                     bstate.capacity = std::numeric_limits<float>::quiet_NaN();
                     bstate.design_capacity = std::numeric_limits<float>::quiet_NaN();
-                    if (bstate.voltage < 22.0) { // 18V = 0%,  22V = 12%
-                        bstate.percentage = std::max(0.0, 0.02 * (bstate.voltage - 18.0));
+                    if (bstate.voltage < 20.0) 
+                      bstate.percentage = 0.0;
                     }
-                    else { // 22V = 12%, 24V = 42%, 26V = 72%. 27V = 87%
-                        bstate.percentage = std::min(1.0, 0.12 + (bstate.voltage - 22.0) * 0.15);
+                    else {
+                        bstate.percentage = std::min(1.0, (bstate.voltage - 20.0) * 0.125);
                     }
                     bstate.power_supply_status = sensor_msgs::BatteryState::POWER_SUPPLY_STATUS_UNKNOWN;
                     bstate.power_supply_health = sensor_msgs::BatteryState::POWER_SUPPLY_HEALTH_UNKNOWN;
--- a/src/motor_hardware.cc
+++ b/src/motor_hardware.cc
@@ -153,12 +153,7 @@ void MotorHardware::readInputs() {
                     bstate.charge = std::numeric_limits<float>::quiet_NaN();
                     bstate.capacity = std::numeric_limits<float>::quiet_NaN();
                     bstate.design_capacity = std::numeric_limits<float>::quiet_NaN();
-                    if (bstate.voltage < 20.0) 
-                      bstate.percentage = 0.0;
-                    }
-                    else {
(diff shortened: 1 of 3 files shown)


src/motor_hardware.cc
```diff
@@ -142,6 +144,22 @@ void MotorHardware::readInputs() {
...
+                case MotorMessage::REG_BATTERY_VOLTAGE: {
+                    int32_t data = mm.getData();
+                    sensor_msgs::BatteryState bstate;
+                    bstate.voltage = (float)data * pid_params.battery_voltage_multiplier +
+                                   pid_params.battery_voltage_offset;
+                    bstate.current = std::numeric_limits<float>::quiet_NaN();
+                    bstate.charge = std::numeric_limits<float>::quiet_NaN();
+                    bstate.capacity = std::numeric_limits<float>::quiet_NaN();
+                    bstate.design_capacity = std::numeric_limits<float>::quiet_NaN();
+                    bstate.percentage = std::numeric_limits<float>::quiet_NaN();
```
rohbotics: Can we put in a dumb percentage that goes between 0-100% between 22-28 volts ?

rohbotics (approved):
Good, but see comments

jim-v:
@rohbotics, how is this?

src/motor_hardware.cc
```diff
@@ -142,6 +144,27 @@ void MotorHardware::readInputs() {
...
+                    int32_t data = mm.getData();
+                    sensor_msgs::BatteryState bstate;
+                    bstate.voltage = (float)data * pid_params.battery_voltage_multiplier +
+                                   pid_params.battery_voltage_offset;
+                    bstate.current = std::numeric_limits<float>::quiet_NaN();
+                    bstate.charge = std::numeric_limits<float>::quiet_NaN();
+                    bstate.capacity = std::numeric_limits<float>::quiet_NaN();
+                    bstate.design_capacity = std::numeric_limits<float>::quiet_NaN();
+                    if (bstate.voltage < 22.0) { // 18V = 0%,  22V = 12%
+                        bstate.percentage = std::max(0.0, 2.0 * (bstate.voltage - 18.0));
```
rohbotics: `bstate.percentage` should be between 0 and 1. `0.02 * (bstate.voltage - 18.0)` should make it work

src/motor_hardware.cc
```diff
@@ -142,6 +144,27 @@ void MotorHardware::readInputs() {
...
+                    }
+                    else { // 22V = 12%, 24V = 42%, 26V = 72%. 27V = 87%
+                        bstate.percentage = std::min(100.0, 12.0 + (bstate.voltage - 22.0) * 15.0);
```
rohbotics: Same issue, should be something like `std::min(1.0, 0.12 + (bstate.voltage - 22.0) * 0.15);`

rohbotics (changes requested):

src/motor_hardware.cc
rohbotics: Also I think our lower voltage (0% level) should be 20V. at 18 you are at 9V per battery which is pretty low.

jim-v:
@rohbotics, good point.  I made it a simple linear function with 20V -> 0 and 28V -> 1

src/motor_hardware.cc
```diff
@@ -142,6 +144,27 @@ void MotorHardware::readInputs() {
...
+                    int32_t data = mm.getData();
+                    sensor_msgs::BatteryState bstate;
+                    bstate.voltage = (float)data * pid_params.battery_voltage_multiplier +
+                                   pid_params.battery_voltage_offset;
+                    bstate.current = std::numeric_limits<float>::quiet_NaN();
+                    bstate.charge = std::numeric_limits<float>::quiet_NaN();
+                    bstate.capacity = std::numeric_limits<float>::quiet_NaN();
+                    bstate.design_capacity = std::numeric_limits<float>::quiet_NaN();
+                    if (bstate.voltage < 20.0) 
+                      bstate.percentage = 0.0;
```
rohbotics: Why use an if, when you could just use std::max(0.0, bstate.percentage), it would be cleaner I think.

rohbotics (changes requested):

rohbotics (approved):
LGTM

PR #35, opened 2018-01-18
Add Program for getting robot info

rohbotics:
I don't have hardware to test on (with the new firmware), so I used a emulated serial port.


jim-v (approved):
I haven't tested (don't have appropriate hardware), but it looks reasonable.

PR #36, opened 2018-02-25
Added support for changing baud rate in probe robot

rohbotics:


src/probe_robot.cc
```diff
@@ -58,6 +59,16 @@ std::string get_option(const std::vector<std::string> &args, const std::string &
     }
 }
 
+int get_option(const std::vector<std::string> &args, const std::string &option,
+                       const int default_val) {
+    auto opt = std::find(args.cbegin(), args.cend(), option);
+    if (opt != args.cend() && opt++ != args.cend()) {
+        return std::stoi(*(opt++));
```
jim-v: Not sure what the benefit of the post-increment of `opt` is on returning from the function.

jim-v (approved):

src/probe_robot.cc
rohbotics: That was a gnarly bug possibility you just caught, the code only worked because the guard above it also used a post increment (of course this means that the guard was broken).

Thanks for catching that!

Fixed it now, along with some better error handling. Please review again.

jim-v (approved):
It looks OK. One thing I noticed is that it doesn't catch unexpected command line arguments.  
```
rosrun ubiquity_motor probe_robot -X
```
It would be nice if it printed the usage message if any unexpected arguments were given.

PR #47, opened 2018-11-04
Keep trying to open serial and communicate

### LINKED ISSUE #43: More graceful shutdown / retry on serial failure
From @mjstn's email

If I start with motor controller serial lines not connected motor_node 
crashes.

Personally I feel we should keep trying ... forever and indicate serial 
not up in ROS error messages.

Anyway, the motor_node crashes.    What was the logic or has it just 
never been made more robust?

rohbotics:
Don'd die if the serial port doesn't exist, or if the firmware isn't talking.

Fixes #43

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(ubiquity_motor)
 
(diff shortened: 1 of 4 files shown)


jim-v (approved):
Looks good, but I haven't tested it.

PR #50, opened 2018-11-11
Added firmware loading tool

### LINKED ISSUE #48: Tool to upgrade firmware
Should be able to download and load the latest firmware.

rohbotics:
Fixes #48

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,10 +118,10 @@ target_link_libraries(probe_robot
 
 ## Mark executable scripts (Python etc.) for installation
 ## in contrast to setup.py, you can choose the destination
-# install(PROGRAMS
-#   scripts/my_python_script
-#   DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
-# )
+install(PROGRAMS
+  scripts/upgrade_firmware.py
+  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
 
 ## Mark executables and/or libraries for installation
 install(TARGETS ubiquity_motor motor_node probe_robot
--- a/package.xml
+++ b/package.xml
@@ -20,5 +20,6 @@
   <depend>dynamic_reconfigure</depend>
   <exec_depend>diff_drive_controller</exec_depend>
   <exec_depend>joint_state_controller</exec_depend>
+  <exec_depend>python-requests</exec_depend>
 
 </package>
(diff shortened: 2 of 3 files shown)


mjstn:
Use of 'python upgrade_firmware.py   results in missing module no matter how I run it in python

ubuntu@ubiquityrobot:~$ rosrun ubiquity_motor probe_robot -f
29
ubuntu@ubiquityrobot:~$
ubuntu@ubiquityrobot:~$ vi upgrade_firmware.py
ubuntu@ubiquityrobot:~$ python upgrade_firmware.py
Traceback (most recent call last):
  File "upgrade_firmware.py", line 2, in <module>
    import requests
ImportError: No module named requests
ubuntu@ubiquityrobot:~$ sudo python ./upgrade_firmware.py
Traceback (most recent call last):
  File "./upgrade_firmware.py", line 2, in <module>
    import requests
ImportError: No module named requests
ubuntu@ubiquityrobot:~$ python3 upgrade_firmware.py
  File "upgrade_firmware.py", line 16
    """
      ^
SyntaxError: Missing parentheses in call to 'print'
ubuntu@ubiquityrobot:~$

rohbotics:
Ah, try running `sudo apt install python-requests`. I will fix this by making it a dependency in the released version.

mjstn:
If we make this available to users on current images we need to tell them to do this install too of course.

If possible we should check if magni-base is active and EXIT at that point.
I say this because a firmware upgrade always has a potential to brick the PSoC so whatever we can do to be checking OURSELVES is more or less 'required'.

As it is now it makes mention about being sure motor-base is stopped but many 'just do it' people may not know what that means and just 'go for it'

Anyway, I exited and stopped magni-base then ran as normal user and it worked and shows up in probe_robot

A single line separated by spaces just after access token is entered should read

Updating firmware now.   Do not power off the robot.  This is expected to take less than  a minute.

rohbotics:
I added it to the rosdep, so when users install from apt they get all the dependencies. 

Also added the print statement you suggested. 

The firmware loading process does exit if magni-base is running, because the loading handshake never completes, but I agree more robust checking is desirable.

rohbotics:
@mjstn I added a check to make sure the serial port is not open.

@jim-v Ignore the travis, it is just borked. I fixed it in the main branch.

jim-v (approved):
Looks OK, but I would prefer to see blank lines between Python functions.

PR #53, opened 2019-02-26
Add several ROS parameters for support of new firmware rev 32 registers

mjstn:
Add ROS parameters that will get pushed to the controller board.


src/motor_node.cc
```diff
@@ -113,14 +113,24 @@ int main(int argc, char* argv[]) {
...
             robot->readInputs();
             r.sleep();
             times++;
         }
     }
 
+    // Tell the controller board firmware what version the hardware is at this time.
+    // TODO: Read from I2C.   At this time we only allow setting the version from ros parameters
+    if (robot->firmware_version >= 32) {
+        ROS_ERROR("DEBUG: Firmware is version %d. Setting Controller board version to %d",
```
rohbotics: This should be ROS_DEBUG and not have the DEBUG in the print

src/motor_node.cc
```diff
@@ -113,14 +113,24 @@ int main(int argc, char* argv[]) {
...
         }
     }
 
+    // Tell the controller board firmware what version the hardware is at this time.
+    // TODO: Read from I2C.   At this time we only allow setting the version from ros parameters
+    if (robot->firmware_version >= 32) {
+        ROS_ERROR("DEBUG: Firmware is version %d. Setting Controller board version to %d", 
+            robot->firmware_version, firmware_params.controller_board_version);
+        robot->setHardwareVersion(firmware_params.controller_board_version);
+        ROS_ERROR("DEBUG: Controller board version has been set to %d",
```
rohbotics: Same note here, use ROS_DEBUG

include/ubiquity_motor/motor_message.h
```diff
@@ -114,7 +118,15 @@ class MotorMessage {
...
-        REG_ROBOT_ID = 0x31,
+        REG_ROBOT_ID = 0x31,	    // Indicates 0 for Magni controller or 1 for Loki robot controller as of late 2018
+
+        REG_MOT_PWR_ACTIVE = 0x32,  // Readback register for host to know if motor controller thinks motor power is active
+        REG_ESTOP_ENABLE   = 0x33,  // An override that may be set to 0 to force motor controller firmware to NOT use some ESTOP logic
+        REG_PID_MAX_ERROR  = 0x34,  // A value that when non-zero enables motor firmware to limit harsh restarts in position after ESTOP release
+
+        REG_MAX_SPEED_FWD  = 0x35,  // Max forward speed cap in a speed message 
+        REG_MAX_SPEED_REV  = 0x36,  // Max reverse speed cap in a speed message  (This is negative)
+        REG_MAX_PWM        = 0x37,  // A value that when non-zero enables motor firmware to limit harsh restarts in position after ESTOP release
```
rohbotics: I think the comment here is misleading, this is the MAX PWM at all times correct?

src/motor_hardware.cc
```diff
@@ -79,8 +79,18 @@ MotorHardware::MotorHardware(ros::NodeHandle nh, CommsParams serial_params,
     prev_pid_params.pid_derivative = -1;
     prev_pid_params.pid_denominator = -1;
     prev_pid_params.pid_moving_buffer_size = -1;
+    prev_pid_params.max_speed_fwd = -1;
+    prev_pid_params.max_speed_rev = -1;
```
rohbotics: Change the name of this member, pid_params and prev_pid params are incorrected as there are a lot more than just PID values

include/ubiquity_motor/motor_parmeters.h
```diff
@@ -83,6 +102,18 @@ struct FirmwareParams {
...
+            nh, "ubiquity_motor/controller_board_version", controller_board_version);
+        estop_enable = getParamOrDefault(
+            nh, "ubiquity_motor/fw_estop_enable", estop_enable);
+        estop_pid_threshold = getParamOrDefault(
+            nh, "ubiquity_motor/fw_estop_pid_threshold", estop_pid_threshold);
+        max_speed_fwd = getParamOrDefault(
+            nh, "ubiquity_motor/fw_max_speed_fwd", max_speed_fwd);
+        max_speed_rev = getParamOrDefault(
+            nh, "ubiquity_motor/fw_max_speed_rev", max_speed_rev);
+        max_pwm = getParamOrDefault(
```
rohbotics, line 115: I am not sure that we want to expose this value in a way that customers can fiddle with it. 
Is there a hard max in firmware to prevent someone from setting this at 100%?

include/ubiquity_motor/motor_parmeters.h
```diff
@@ -83,6 +102,18 @@ struct FirmwareParams {
             nh, "ubiquity_motor/pid_denominator", pid_denominator);
         pid_moving_buffer_size = getParamOrDefault(
             nh, "ubiquity_motor/window_size", pid_moving_buffer_size);
+        controller_board_version = getParamOrDefault(
+            nh, "ubiquity_motor/controller_board_version", controller_board_version);
+        estop_enable = getParamOrDefault(
+            nh, "ubiquity_motor/fw_estop_enable", estop_enable);
```
rohbotics: This needs a better name than 'estop enable' maybe like estop_detect?

src/motor_hardware.cc
```diff
@@ -202,6 +212,53 @@ void MotorHardware::requestVersion() {
     motor_serial_->transmitCommand(version);
 }
 
+
+// Due to greatly limited pins on the firmware processor the host figures out the hardware rev and sends it to fw
+// The hardware version is 0x0000MMmm  where MM is major rev like 4 and mm is minor rev like 9 for first units.
+// The 1st firmware version this is set for is 32, before it was always 1
+void MotorHardware::setHardwareVersion(int32_t hardware_version) {
+    ROS_ERROR("setting hardware_version to %x", (int)hardware_version);
```
rohbotics: Not a ROS_ERROR, a ROS_INFO at most

src/motor_hardware.cc
```diff
@@ -284,6 +344,59 @@ void MotorHardware::sendParams() {
         commands.push_back(winsize);
     }
 
+    if (cycle == 5 && (pid_params.max_pwm != prev_pid_params.max_pwm)) {
```
rohbotics: All of this seems to be duplicated from the individual functions that you made for these parameters?

We only ever need to send this data once, so I don't see why it needs to be with all this dynamic reconfigure stuff here

rohbotics (changes requested):

include/ubiquity_motor/motor_message.h
mjstn: Wrong comment due to cut-paste.  I will make that comment:
 The maximum wheel driver PWM value that will be used on the motor driver

src/motor_hardware.cc
mjstn: I'll make it prev_firmware_params as that structure holds all firmware params as the class implies using FirmwareParams firmware_params

include/ubiquity_motor/motor_parmeters.h
mjstn: This is in fact an enable of the hardware signal estop logic only available on rev 5.0 boards.  The logic is a bit contorted due to two modes being implemented.   There is a pre rev 5.0 mode that uses the threshold and then there is the signal that shows if switch was setting motor power on or off.    The prefered and not hacky mode is use rev 5.0 signal and the estop_enable register allows us to disable the hardware rev 5.0 mode of estop logic.    I really wish we just wrote off estop on rev 4.9 but sadly we shipped those in units so must kludge it there.

src/motor_hardware.cc
mjstn: Ok but it is modeled after pre-existing deadman timer so I just did the same.  I'll convert to INFO for the new changes.

src/motor_hardware.cc
mjstn: that part of the code confused me basically so I did those to be like the other pid parameters.   These indeed do not need to be dynamically changed so if that is the purpose of code in MotorHardware::sendParams  I'm fine with taking it out and will do so.

src/motor_node.cc
mjstn: That is left over debug on the 2 ROS_ERROR lines so thanks for review.

include/ubiquity_motor/motor_parmeters.h
rohbotics: I understand what it is, the name is just misleading. 

If i make `estop_enable` false then does the estop button on the front no longer do anything? Obviously not.

A name like `estop_detection` would be make it more clear as to what is actually happening

include/ubiquity_motor/motor_parmeters.h
mjstn: We do have a chance to wrong a past lack of speed limits and that may or may not be a good thing.
In the past the firmware sees the 32bit value and  upper 16 bits is one speed, lower 16 is other.  That of course is a possible desaster but was outside the scope of this fix.   This fix is to allow some lower value to be set.  If the user cranks this up to 0xfffe  for each one he is no worse off than the past.   Yes it is really not good but we have not addressed this issue and instead we place a limit in this way.  To revert to old max limits user can set these full on and the masking of 0xffff still is the old limit in that case.

mjstn: You have reminded me to follow up on something with the reverse limit so I'm going to re-inspect that and try some more tests prior to the next review which as you say is for me to simply push to this branch on remote and you will see update on the pull request.

include/ubiquity_motor/motor_parmeters.h
rohbotics: Mark can you take a look at this again

mjstn: Ok, it is pushed.  Tell me if you don't see it.

rohbotics (approved):
Thanks for the updates, mergeing

PR #55, opened 2019-03-27
Allow selecting what version to download

### LINKED ISSUE #54: Firmware update utility should allow going to previous version
Some way to tell the firmware updater which version to load.

rohbotics:
Fixes #54

### DIFF
--- a/scripts/upgrade_firmware.py
+++ b/scripts/upgrade_firmware.py
@@ -24,18 +24,24 @@ if (subprocess.call(['fuser','-v', '/dev/ttyAMA0'], stdout=None) == 0):
 
 email = raw_input("Please enter your email address: ").strip()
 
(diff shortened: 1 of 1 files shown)


rohbotics:
Hey @mjstn 

Can you take a quick look at this? It should solve the roll back issue you were talking about.

mjstn (approved):
This will do the trick.   I will discuss later with you what I think I will ask Joe to add to our firmware-upgrade learn page that will then document this and perhaps mention a few key very recent revision numbers as a sort of mini hint of numbers that will only be shown if those revs would be supported by this tool where it would only show the last one we know is wide circulation and a good release.

rohbotics:
Thanks for the review @mjstn 

Merging

PR #61, opened 2019-04-16
Enhance ESTOP recovery. Publish motor power state on topic. Clarify a Deserialization error

mjstn:
- Make estop state is known to the motor node.
- Publish motor power state in /diagnostics topic
- Switch to ROS_WARN for special case of unknown (new) register seen from firmware
- Send speeds of 0 to controller when in ESTOP and for first fraction of a second after ESTOP release

### DIFF
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -96,7 +96,8 @@ public:
     virtual ~MotorHardware();
     void clearCommands();
     void readInputs();
-    void writeSpeeds(bool zeroSpeeds);
+    void writeSpeeds();
+    void writeSpeedsWithDisable(bool disableSpeeds);
     void requestVersion();
     void setParams(FirmwareParams firmware_params);
     void sendParams();
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -97,7 +97,7 @@ public:
     void clearCommands();
     void readInputs();
     void writeSpeeds();
-    void writeSpeedsWithDisable(bool disableSpeeds);
+    void writeSpeedsInRadians(double  left_radians, double  right_radians);
     void requestVersion();
     void setParams(FirmwareParams firmware_params);
     void sendParams();
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -65,6 +65,11 @@ struct MotorDiagnostics {
     bool right_pwm_limit = false;
     bool left_integral_limit = false;
     bool right_integral_limit = false;
+    bool left_max_speed_limit = false;
+    bool right_max_speed_limit = false;
+
+    // Motor power is inactive, most likely from ESTOP switch
+    bool estop_motor_power_off = false;
 
     // Power supply statuses
     float battery_voltage = 0.0;
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -68,9 +68,6 @@ struct MotorDiagnostics {
     bool left_max_speed_limit = false;
     bool right_max_speed_limit = false;
 
-    // Motor power is inactive, most likely from ESTOP switch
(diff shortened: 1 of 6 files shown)


include/ubiquity_motor/motor_hardware.h
```diff
@@ -91,7 +96,7 @@ class MotorHardware : public hardware_interface::RobotHW {
     virtual ~MotorHardware();
     void clearCommands();
     void readInputs();
-    void writeSpeeds();
+    void writeSpeeds(bool zeroSpeeds);
```
rohbotics: I don't like this function signature, when you are calling it with `writeSpeeds(true)` it is completely unclear what the true or false means. In fact it looks more like true would mean actually sending the speeds.

I would make a seperate function called `zeroSpeeds` or `stopRobot` where it is very obvious from the call site what is happening.

include/ubiquity_motor/motor_message.h
```diff
@@ -159,6 +164,11 @@ class MotorMessage {
     RawMotorMessage serialize() const;
 
     int deserialize(const RawMotorMessage &serialized);
+    const static int ERR_DELIMITER        = 1;
```
rohbotics: This should really be an enum class that deserialize returns.

src/motor_hardware.cc
```diff
@@ -190,19 +206,47 @@ void MotorHardware::readInputs() {
...
     MotorMessage both;
     both.setRegister(MotorMessage::REG_BOTH_SPEED_SET);
     both.setType(MotorMessage::TYPE_WRITE);
-    int16_t left_tics = calculateTicsFromRadians(joints_[0].velocity_command);
-    int16_t right_tics = calculateTicsFromRadians(joints_[1].velocity_command);
+    double  left_radians = joints_[0].velocity_command;
+    double  right_radians = joints_[1].velocity_command;
+
+    if (zeroSpeeds) {    // Force sending of zero speeds as an override
+        left_radians  = (double)(0.0);
```
rohbotics: No need to cast to double here, that is the default.

src/motor_message.cc
```diff
@@ -155,19 +155,20 @@ int MotorMessage::deserialize(const RawMotorMessage &serialized) {
                                   serialized.begin() + 7, data.begin());
                         return 0;
                     } else
-                        return 5;
+                        return MotorMessage::ERR_UNKNOWN_REGISTER;
```
rohbotics, line 158: Again enum class would be the cleanest

src/motor_node.cc
```diff
@@ -163,6 +163,9 @@ int main(int argc, char* argv[]) {
     last_time = ros::Time::now();
     ROS_WARN("Starting motor control node now");
 
+    // Implement a speed reset while ESTOP is active and a delay after release
+    double estopReleaseDeadtime = (double)(0.8);
```
rohbotics: again please remove the unnecessary and noisy casts

src/motor_node.cc
```diff
@@ -180,7 +183,20 @@ int main(int argc, char* argv[]) {
...
-        robot->writeSpeeds();
+
+        // Update motor controller speeds.
+        if (robot->getEstopState()) {
+            robot->writeSpeeds(true);    // We send zero velocity when estop is active
+            estopReleaseDelay = estopReleaseDeadtime;
+        } else {
+            if (estopReleaseDelay > (double)(0.0)) {
+                // Implement a delay after estop release where velocity remains zero
+                estopReleaseDelay -= ((double)(1.0)/node_params.controller_loop_rate);
```
rohbotics: casts.

rohbotics (changes requested):
A bunch of mostly code style things.

I can implement these if you wish, will only take a couple minutes.

rohbotics:
Another note: Try to always run with tests `catkin_make run_tests` that way you will know if you broke something that we expect to work. Sometimes they catch actual bugs. In this case it was just a compile time thing.

rohbotics (approved):

rohbotics:
Hey @mjstn I went ahead and made the minor changes that I wanted.

PR #62, opened 2019-04-17
Publish estop state as a bool

### LINKED ISSUE #60: Publish E-Stop State

rohbotics:
Needs testing

Fixes #60

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ set(CMAKE_CXX_STANDARD 14) # use C++14
 find_package(catkin REQUIRED COMPONENTS 
 	roscpp 
 	geometry_msgs  
+	std_msgs  
 	nav_msgs 
 	serial 
 	hardware_interface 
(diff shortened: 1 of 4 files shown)


mjstn (approved):
This deals with the current need for a simple and easily consumable topic that shows motor power state.
Ideally a topic with motor_state and then motor_power under it would be extensible for future needs but this will be ok I think so I will approve as it is a deliverable in some form that a customer is demanding.

mjstn (commented):
Oh crud.   Notice that motor power active is true if motor power is on.
This means that motor power active is actually the inversion of estop active.
The diagnostic message now does show motor power state properly using an inversion.

rohbotics:
Thanks for catching the inversion problem, fixed.

jim-v (approved):

mjstn (approved):
Looks good now.  I have switched to this branch and tested the code here and it follows the actual switch properly in state.

PR #67, opened 2019-05-03
Allow passing in serial port for non-standard setups

### LINKED ISSUE #64: Configurable serial port name for firmware downloader
Optionally read it directly from the yaml file

rohbotics:
Useful for people using SBCs other than Pis, or FTDIs

Closes #64


mjstn (approved):
This looks good.   Thank you for doing this to help those not using RasPi OR persons like me who use dev/ttyUSB0 intentially so they can power on and off the controller without reboot of pi.

PR #68, opened 2019-05-04
Adds support for host to know if firmware is using options tied to th…

mjstn:
Enables host to see hardware options from a new message from firmware.  
Host can now detect option that hardware is using 6-state encoders and then this host code adjusts odometry for proper tics to radians values so that odometry is correct.
Lastly we change the warning for unrecognized message from firmware so it is now a ROS_WARN_ONCE.  Meant to do this last time but did not do it then.
This host code will support legacy firmware or newer 6-state encoder enabled firmware.


include/ubiquity_motor/motor_parmeters.h
```diff
@@ -116,6 +119,8 @@ struct FirmwareParams {
             nh, "ubiquity_motor/fw_max_pwm", max_pwm);
         deadman_timer = getParamOrDefault(
             nh, "ubiquity_motor/deadman_timer", deadman_timer);
+        hw_options = getParamOrDefault(
```
rohbotics: I strongly object to this being a parameter, it should *only* ever be read from the firmware

src/motor_hardware.cc
```diff
@@ -38,11 +38,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 // 17.2328767123
 // gear ratio of 4.29411764706:1
-#define TICS_PER_RADIAN (41.0058030317 / 2)
-#define QTICS_PER_RADIAN (TICS_PER_RADIAN * 4)
+#define TICS_PER_RADIAN_ENC_3_STATE (41.0058030317/2)
+#define QTICS_PER_RADIAN_ENC_3_STATE (tics_per_radian*4)
 #define VELOCITY_READ_PER_SECOND \
     10.0  // read = ticks / (100 ms), so we have scale of 10 for ticks/second
-#define CURRENT_FIRMWARE_VERSION 24
+#define LOWEST_FIRMWARE_VERSION 28
```
rohbotics, line 49: Thanks for updating this

rohbotics (changes requested):

include/ubiquity_motor/motor_parmeters.h
mjstn: Agreed.  Not sure why I had it as settable from ROS side but at the time I was thinking we may want to make it switchable in some way which we discussed. I had not gone back and removed these 2 lines after that.   It does seem that it is a sort of read only member of FirmwareParams and so I leave it there so it is tightly bound to firmware, that part seems right.  The option we had discussed on the phone was if we ever wanted to try to make it switchable and I am glad you agreed we do not as that would be problematic.

rohbotics:
@mjstn Is this ready for re-review?

jim-v (approved):
:+1:

rohbotics (approved):

PR #76, opened 2019-07-21
Added --file option

### LINKED ISSUE #74: Support for loading firmware from a file
If we give a customer a special firmware file, we need them to be able to load it without downloading any internal utilities. 

https://github.com/UbiquityRobotics/ubiquity_motor/blob/kinetic-devel/scripts/upgrade_firmware.py needs to take a command line argument called "--file" that accepts a path to a firmware file and then bypasses the download step to start loading the file directly.

asergeenko:
Added command line argument called "--file" that accepts a path to a firmware file and then bypasses the download step to start loading the file directly

https://github.com/UbiquityRobotics/ubiquity_motor/issues/74

Fix #74

### DIFF
--- a/scripts/upgrade_firmware.py
+++ b/scripts/upgrade_firmware.py
@@ -17,8 +17,12 @@ Note: Updating the firmware requires access to the internet.
 -------------------------------------------------------------
 """
 
+TMP_FILE_PATH = '/tmp/firmware'
+
+
 parser = argparse.ArgumentParser(description='Ubiquity Robotics Firmware Updater')
 parser.add_argument('--device', help='Serial port to use (eg /dev/ttyAMA0)', default='/dev/ttyAMA0')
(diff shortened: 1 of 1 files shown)


scripts/upgrade_firmware.py
```diff
@@ -17,8 +17,12 @@
 -------------------------------------------------------------
 """
 
+TMP_FILE_PATH = '/tmp/firmware'
```
rohbotics, line 20: Making it a constant definitely cleaner

rohbotics (approved):
Thanks for the PR, it looks good to me :+1:

rohbotics:
@mjstn If you pull the latest ubiquity_motor you should be able to use `rosrun ubiquity_motor upgrade_firmware.py --file PATH` to load a file from disk. This should replace the internal firmware_downloader once it is released.

PR #77, opened 2019-07-23
I75

asergeenko:


rohbotics:
Somehow the other commit made it into this PR. Please rebase to fix this.

scripts/test_velocity.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
+
+# VELOCITY can be positive (driving forward) or negative (driving backward)
+VELOCITY = 0.05
```
rohbotics: This is slower than the robot can really go, lets change it to 0.2 m/s.

scripts/test_velocity.py
```diff
@@ -0,0 +1,53 @@
...
+
+# Initial turn angle (Z axis)
+ANGLE = 0.0
+
+import roslib; roslib.load_manifest('ubiquity_motor')
```
rohbotics: This is not needed in the newer versions of ROS that we use.

scripts/test_velocity.py
```diff
@@ -0,0 +1,53 @@
...
+    while not rospy.is_shutdown():
+
+        # Publish the velocity
+        pub.publish(vel_msg)
+
+        if initial:
+            vel_msg.angular.z = 0.0
+            initial = False
+        
+        rospy.sleep(0.001)
```
rohbotics: 20hz is fast enough for something like this.

scripts/test_velocity.py
```diff
@@ -0,0 +1,53 @@
...
+print ("""-------------------------------------------------------------
+Odometer consistency check
+-------------------------------------------------------------
+""")
+
+
+def odometry_callback(msg):
+
+    # Calculate velocity error (%)
+    print ("Velocity error: {}%".format(round(abs(msg.twist.twist.linear.x-VELOCITY)/VELOCITY*100,2)))
```
rohbotics: It is also useful to check position error, make sure that the change in position is the command velocity integrated with respect to time.

rohbotics (changes requested):
Hi Alexsandr, overall this is pretty good for a first ROS thing, just a couple small things to change.

Let me know if you have any questions.

rohbotics (approved):

rohbotics:
This looks good to me.

PR #81, opened 2020-03-19
Better error messages for common errors

### LINKED ISSUE #238: upgrade_firmware.py  wrong access token error error give no hint at what is wrong
When an incorrect access token is entered in upgrade_firmware.py the script accepts it which leads the user to think the code was ok.
Then a version is entered and right away this error comes up

Error downloading firmware 401

A real geek may realize this is the 'Unauthorized' error but it is going to lead to confusion if we do not explain it in human readable form.

To help users move on we should help by saying it is error 401 like above but also then say  
'unauthorized access most likely due to incorrect or timed out token entry'.
The access tokens are only good for one hour from the time of reception.

rohbotics:
Fixes https://github.com/UbiquityRobotics/ubiquity_main/issues/238

### DIFF
--- a/scripts/upgrade_firmware.py
+++ b/scripts/upgrade_firmware.py
@@ -45,7 +45,9 @@ else:
     if email != "":
         r = requests.post('https://api.ubiquityrobotics.com/token/', json = {'email': email}) 
 
-        if r.status_code != 200:
+        if r.status_code == 500:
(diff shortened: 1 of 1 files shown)


rohbotics:
@mjstn Can you review this fairly quick change?

mjstn (approved):
This will help customers and looks great

PR #84, opened 2020-06-08
Mcb v37 changes

mjstn:
A group of changes to do the following

- Add new HW register names for registers created for v36 and v37
- Add support routines to request these registers and get values as an inbound read comes back
- Add set routine so host can set new registers as required
- Add recognition of a wheel_type base.yaml parameter that can set wheel type to 'standard' or 'thin' and defaults to 'standard'.  This gets converted to a bit setting in the MCB.
- Added powerup query of wheel type and print that in ros log for reference
- Using above read the 8-bit option/board rev switch on node startup and echo it back to a new MCB register that holds the option switch values.  This is added so firmware can 'see' this 8-bit I2C port.
- Newer firmware has a system event register which right now indicates if the MCB has been reset. This change adds ability to monitor if MCB was reset and notify in ros log.  In a different change we will act on this but for now it at least gives us ability to monitor MCB resets.
- Add a base.yaml parameter of  fw_max_pwm so user can tweek max_pwm both from base.yaml and from dynamic parameters.
- Fix some log spam issues for firmware version and date by changing to ONCE type of logs for those.


include/ubiquity_motor/motor_hardware.h
```diff
@@ -45,9 +45,16 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-#include <ubiquity_motor/motor_parmeters.h>
+#include <ubiquity_motor/motor_parameters.h>
 #include <ubiquity_motor/motor_serial.h>
 
+// To access I2C we need some system includes
+#include <linux/i2c-dev.h>
+#include <fcntl.h>
+#include <sys/ioctl.h>
+#define  I2C_DEVICE  "/dev/i2c-1"     // This is specific to default Magni I2C port on host
+#define  I2C_PCF8574_8BIT_ADDR  (uint8_t)0x40  // I2C addresses are 7 bits but often shown as 8-bit
```
rohbotics: Its easier to just use a constant variable here `const static uint8_t I2C_PCF8574_8BIT_ADDR = 0x40`.  This way the compiler gets the types earlier and can give better error messages if you screw something up. Also under a optimized build, it will be equivalent anyway.

scripts/system_info.py
```diff
@@ -0,0 +1,222 @@
+#!/usr/bin/python
```
rohbotics: This is useful and good, but shouldn't be ubiquity_motor, its not motor related really.

scripts/system_info.py
```diff
@@ -0,0 +1,222 @@
...
+Copyright (c) 2020, Ubiquity Robotics
+All rights reserved.
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+* Neither the name of fiducial_follow nor the names of its
```
rohbotics: Change the package name in this sentence to whatever we decide to put this in.

include/ubiquity_motor/motor_message.h
```diff
@@ -68,80 +75,85 @@ class MotorMessage {
 
     // Registers enum in class to avoid global namespace pollution
     enum Registers {
-        REG_STOP_START = 0x00,
+        REG_STOP_START = 0x00,          // Depricated
```
rohbotics: sp: Deprecated

include/ubiquity_motor/motor_hardware.h
rohbotics: Also does this need to be the header? I these things should be const variables in the implementation file

include/ubiquity_motor/motor_message.h
```diff
@@ -154,9 +166,45 @@ class MotorMessage {
         DEBUG_58 = 0x58
     };
 
+    // PID Control values used in special register PID_CONTROL
+    enum PidControlActions {
+        PID_CTRL_RESET               = 1,
+        PID_CTRL_PWM_OVERRIDE        = 2,
+        PID_CTRL_P_ONLY_ON_0_VEL     = 4
```
rohbotics: Do we want to support this mode? Does it acutally stop chatter?

src/motor_node.cc
```diff
@@ -138,17 +141,48 @@ int main(int argc, char* argv[]) {
         robot->requestFirmwareDate();
     }
 
+    // Determine hardware options that can be set by the host to override firmware defaults
+    int32_t host_setable_hw_options = 0;
+    if (node_params.wheel_type.compare("standard") == 0) {
```
rohbotics: Just use `==`, this is C++ we have overloads (for better or for worse).

`if (node_params.wheel_type == "standard") {`

src/motor_node.cc
```diff
@@ -138,17 +141,48 @@ int main(int argc, char* argv[]) {
...
+    } else if (node_params.wheel_type.compare("thin")  == 0){
+        host_setable_hw_options |= MotorMessage::OPT_WHEEL_TYPE_THIN;
+        ROS_INFO("Host is specifying hardware wheel_type of '%s'", "thin");
+    } else if (node_params.wheel_type.compare("firmware_default") == 0) {
+        // Here there is no specification so the firmware default will be used
+        ROS_INFO("Firmware default wheel_type will be used.");
+    } else {
+        ROS_WARN("Invalid wheel_type of '%s' specified! Using wheel type of standard", 
+            node_params.wheel_type.c_str());
+        node_params.wheel_type = "standard";
```
rohbotics, line 158: Shouldn't the default be "firmware_default" ?

src/motor_hardware.cc
```diff
@@ -664,12 +764,92 @@ void MotorDiagnostics::motor_power_status(DiagnosticStatusWrapper &stat) {
...
+    if (bytesRead != NumByteToRead) {      // verify the number of bytes we requested were read
+      retCode = -9;
+      goto exitWithFileCloseAndSemUnlock;
+    }
+    retCode = bytesRead;
+
+  exitWithFileCloseAndSemUnlock:
+    close(fd);
+
+  exitWithSemUnlock:
```
rohbotics: These label names don't really make sense, we aren'r using the sem_lock

src/motor_hardware.cc
```diff
@@ -664,12 +764,92 @@ void MotorDiagnostics::motor_power_status(DiagnosticStatusWrapper &stat) {
...
+   // If your system requires a lock on the bus, it should be called here 
+   // if ((semId >= 0) && (ipc_sem_lock(semId) < 0)) {
+   //   // printf("i2c_BufferRead: Cannot obtain I2C lock!  ERROR: %s\n", strerror(errno));
+   //   return -1;
+   // }
+
+    // we are now free to access the I2C hardware
+    int fd;                                         // File descrition
+    int  address   = i2c8bitAddr >> 1;              // Address of the I2C device
+    unsigned char buf[8];                           // Buffer for data being written to the i2c device
```
rohbotics: Use `uint8_t`

src/motor_hardware.cc
```diff
@@ -664,12 +764,92 @@ void MotorDiagnostics::motor_power_status(DiagnosticStatusWrapper &stat) {
...
+
+    if ((fd = open(i2cDevFile, O_RDWR)) < 0) {        // Open port for reading and writing
+      retCode = -2;
+      ROS_ERROR("Cannot open I2C def of %s with error %s", i2cDevFile, strerror(errno));
+      goto exitWithSemUnlock;
+    }
+
+    // The ioctl here will address the I2C slave device making it ready for 1 or more other bytes
+    if (ioctl(fd, I2C_SLAVE, address) < 0) {        // Set the port options and addr of the dev
```
rohbotics: This error comparisons should probably just check for any non-0 return to remove some noise from the code.

`if (ioctl(fd, I2C_SLAVE, address))

src/motor_node.cc
```diff
@@ -183,10 +217,14 @@ int main(int argc, char* argv[]) {
     float minCycleTime = 0.75 * expectedCycleTime;
     float maxCycleTime = 1.25 * expectedCycleTime;
 
+    double sysEventQueryPeriod    = 5.0;
```
rohbotics: Should probably be a ros::Duration, and then you can do all your comparisons as durations instead of converting to seconds.

src/motor_hardware.cc
```diff
@@ -664,12 +764,92 @@ void MotorDiagnostics::motor_power_status(DiagnosticStatusWrapper &stat) {
 // motor_encoder_mode returns 0 for legacy encoders and 1 for 6-state firmware
 void MotorDiagnostics::firmware_options_status(DiagnosticStatusWrapper &stat) {
     stat.add("Firmware Options", firmware_options);
+    std::string option_descriptions("");
     if (firmware_options & MotorMessage::OPT_ENC_6_STATE) {
-        stat.summary(DiagnosticStatusWrapper::OK, "High resolution encoders");
-    } 
-    else {
-        stat.summary(DiagnosticStatusWrapper::OK, "Standard resolution encoders");
+        option_descriptions = option_descriptions + "High resolution encoders";
```
rohbotics: Can just use +=

rohbotics (changes requested):
Ok did a review of this, have a few nits and questions.

Also on a more general note, I would prefer if the PRs were smaller and more frequent. This makes it easier to review, and gets stuff into master faster allowing other people to work on top it.

I would have prefered the system_event stuff, thin wheel support, self test to be separete PRs that were made whenever that part of the code was finished.

include/ubiquity_motor/motor_hardware.h
mjstn: Its funny but I agree that these are best in the implementation file.  I will indeed move those to the src and use constants at the same time.  The history is I have a big I2C file and these were peeled off of that and put here because to pull in the entire thing would have meant a bigger project so I put in just what is required.  Granted not the 'ideal' case but by far the most efficient since I also changed everything to ROS_ logs .

scripts/system_info.py
mjstn: I believe I asked you about this but never followed through.   It belongs in magni robot but knowing that the merge there is so ancient I was concerned it may not get there ... ever.    I am happy to move it BUT want BADLY that this tool, crude as it is, be out in field on a very recent image.  We can discuss and I'll take it out of this pull request

include/ubiquity_motor/motor_message.h
mjstn: This mode is used in test.  It is in firmware now to do the selftest.  So it is here more to sync up and make the bits meaning be seen on host code as well.

src/motor_hardware.cc
mjstn: Another question I had for you.  As I said, this comes from my own code where I used this.  I left inklings of it here so I could ask you about it.    My understanding is in ROS different nodes cannot be running at the same time so a lock for I2C is sort of automatic in ROS.   If you feel that is a true statement I'll take out all the lock stuff, even the bits in commented out source.     Again, I meant to ask about this so 'there you are'.

include/ubiquity_motor/motor_message.h
mjstn: The chatter bug is not fixed yet so I will take this out at this time.

src/motor_node.cc
rohbotics: @mjstn

scripts/system_info.py
rohbotics: @mjstn

src/motor_hardware.cc
rohbotics: The whole point of ROS is that you do have multiple nodes operating in parallel. Within the same node, the threading model is up to you. I will have to think about the i2c stuff a bit more, but we can do without a lock for now (since nobody else is currently using i2c along side us).

src/motor_node.cc
rohbotics: @mjstn

src/motor_hardware.cc
rohbotics: Looking into this some more. the kernel provides a cleaner atomic interface for a I2C register read operation using the I2C_RDWR ioctl. This does all the reading and writing in one step atomically, so that no other processes or kernel modules can interrupt it.

We should use this functionality. As it stands this code is brittle to other systems trying to use I2C at the same time (including the kernel RTC module).

See I2C_RDWR in here https://www.kernel.org/doc/Documentation/i2c/dev-interface
And some example code here: https://gist.github.com/JamesDunne/9b7fbedb74c22ccc833059623f47beb7

scripts/system_info.py
mjstn: this has moved to magni_bringup/scripts and is no longer a part of this

src/motor_node.cc
mjstn: Here we get into a bit of 'personal choice'.   If nothing is specified, firmware default.  But if somebody from HOST wants to set it but messed up we should right there tell them they made incorrect choice and set it to what the host side feels is the default.   Otherwise it gets really messy as we have to query firmware which is not really up yet and find setting and translate it and show that it will be  blah blah.   I say if host side tries to set it, force a host side defult.

src/motor_hardware.cc
mjstn: I have changed the labels and taken out lock stuff.  We will approach the lock issue as a separate change and do as you suggest at that time.

src/motor_node.cc
mjstn: Ok.  I don't really like that verbose way either, it is just more 'formal'.  Glad to go to 'normal' and readable format any day.

rohbotics (approved):

PR #87, opened 2020-06-19
Fixes ODOM bug and modifies way we set wheel type to safer system.

mjstn:
- The REG_HW_OPTIONS register becomes read only so the write support to that reg is removed.
- A new setWheelType interface is added to be able to set  'thin' or 'standard' wheel_type in yaml file.

These things allow full support for wheel type AND fix bug for ODOM introduced by earlier attempt to support thin wheels.


mjstn:
Darn, I never hit the actual 'Request' button till just now.  This has been ready for over a day.  Nuts.

rohbotics:
@mjstn There is a merge conflict. Please rebase your changes on top of the master branch and resolve the conflicts.

mjstn:
have compared what is now in mayV37Changes with my earlier copy of full ubiquity_motor and they look to be same.
Thank you for working on sorting out the 'evil mess' I made.   It is a lesson I hope to remember.

So is this then ready for review / approval as next step and is able to be merged OR is there a need still to form a totally clean version from a fresh and pulled kinetic-devel?   I will do that if it is still required and use totally different branch name of course.

rohbotics (approved):
Make sure that the change still has all the components that you wanted it to have, but what is here looks good.

mjstn:
Finally merging this.  I will NEVER EVER again use a branch twice for two pull requests!!!!!!

Issue #83, opened 2020-05-12
Robot does not move "Baud rate issue"


kristjanp28:
Problem: Robot dosen't move with teleop.
Procedure for reproduction the error:

MCB: 5.2 known working board.
Image: 2020-02-10-ubiquity-xenial-lxde-raspberry-pi.img
Firmware: v35

procedure:
- Power up
- wait for SIN and SOUT to blink / 3 minutes to boot up
- move robot around with teleop
- shutdown with 'sudo shutdown -h now'
- power off with switch

observations:
- motor_node uses ~60% of cpu
- `sudo stty -F /dev/ttyAMA0` return baud rate of 38400
- rosout.log returns bunch of 'REJECT' messages
- killing motor_node and restarting it with `rosrun ubiquity_motor motor_node` returns exact 'REJECT' messages as log file.
- setting baud rate to 38400 with `sudo stty -F /dev/ttyAMA0 38400` and starting the motor_node returns no errors.

rohbotics:
@mjstn @kristjanp28 Have we gotten this bug again?

mjstn:
This defect is a duplicate of https://github.com/UbiquityRobotics/ubiquity_motor/issues/95
Closing this now as it is the same defect.

PR #88, opened 2020-06-21
Workaround for MCB baud rate bug

mjstn:
This is an attempt to workaround ubiquity_motor issue #83 

We felt that some linux piece of software is initializing the /dev/ttyAMA0 serial port to 115200 baud sometimes AFTER we set it up in our motor node.

This workaround holds off on instantiation of the motor_serial instance for a couple seconds.

Although it does not really FIX root cause, it does seem to be a highly effective workaround.


mjstn:
Accidental pull request from wrong branch.  Closing

PR #94, opened 2020-07-19
Add mcb message holdoff on host side until MCB indicates no longer busy

mjstn:
a new motor_message.h bit called MCB_BUSY_STATE_MASK is allowed to be controlled by firmware.
Host will set a busy state in motor_hardware.cc when sending queries to MCB.
MCB must return any message with this bit cleared (was upper bit of mcb register which we never have used anyway).

The intent of this change is to be backward compatible with existing firmware and allow future firmware such as v38 to extend any MCB busy time within reason as required.

The change as it exists now will give up and send messages if it takes over 200ms so we do not break things badly.
That is a little 'dirty' but frankly for this sort of architectural change so late in the stages of assorted firmware this is cautious.

### DIFF
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -155,7 +155,8 @@ private:
     FirmwareParams fw_params;
     FirmwareParams prev_fw_params;
 
-    int32_t mcbIsBusyNow;
+    uint8_t mcbIsBusyNow;
+    uint8_t mcbLastReadReg;
 
     int32_t deadman_timer;
 
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -127,6 +127,7 @@ public:
     int  getOptionSwitch(void);
     void setOptionSwitchReg(int32_t option_switch);
     void requestSystemEvents();
+    float getBatteryVoltage();
     void setSystemEvents(int32_t system_events);
     int firmware_version;
     int firmware_date;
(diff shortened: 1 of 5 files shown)


rohbotics (changes requested):
Thanks for putting work into this issue, it will really help make our motor controller communication more robust.
Before I get to the nitty gritty of the review, I have some comments of the design of this feature.

1) I think this will reduce the register space space by half, correct? If we ever want to have more than 128 registers, this will lock us out of it. We already barely care about the protocol version field, could we cut that down to 3 bits and use the high bit for busy state instead?

2) I think the tracking of MCB busy states should be encapsulated in motor_serial, and not in motor_hardware. Basically make it so that MotorSerial::transmitCommand does what MotorHardware::transmitMcbCommand does in this PR.

mjstn:
Where the bit resides is a topic of study.  That I will reply to later.

Where the logic resides seems higher level than serial.  serial is a packet driver  that should not know or try to think about internal mcb state is my thought.     If the busy bit has anything to do with the content of the message, which I can see cases for, then an automatic setting of this state in a packet driver seems not a good place to manage the bit.  Notice that the busy bit is only sometimes set (for queries) so we do not overrun MCB on queries.  I have not yet determined if there are some cases for writes where we want to also set busy bit.     Also the idea is the mcb itself can set busy bit which is a firmware decision that adds flexibility for firmware to try to hold off host layer messages.  So having a layer that is the only interface on top of the mcb that can understand what messages require what busy bit interactions is why it is in motor_hardware.cc

mjstn:
This will go off of review and will remain in the branch until such time a scheme is agreed upon for this issue.

PR #97, opened 2020-08-03
Fix wheel direction issues

mjstn:
This is the host side support for fix of every 15 sec short speed dropout.
Refer to: https://github.com/UbiquityRobotics/avalon/issues/24

This also adds a wheel speed reversal feature new to v38 firmware


rohbotics:
Is this ready for review? It still uses a bunch of the controllerLooprate sleeps, so it will cause a bunch of problems right?

mjstn:
This is now ready for review.  I should have thought about that bit of cleanup use the same mechanism for all the earlier messages.   The general idea is we make a system that deals with avoiding overruns.   This code today is to avoid a situation we knew was leading to issues.   Also this code adds some small enhancement to allow for checking up on the MCB state by the host code.

rohbotics (approved):
This still doesn't fix any of the ctrlLoopDelay issues. I am expecting a separate PR for that. The wheel direction fixes in here can be merged.

mjstn:
Will do pull request again after merge from master seems in order.

PR #99, opened 2020-08-07
Fix15sec deadman hickup

mjstn:
Remove use of improper delays

### DIFF
--- a/src/motor_hardware.cc
+++ b/src/motor_hardware.cc
@@ -87,7 +87,7 @@ MotorHardware::MotorHardware(ros::NodeHandle nh, CommsParams serial_params,
     // Insert a delay prior to serial port setup to avoid a race defect.
     // We see soemtimes the OS sets the port to 115200 baud just after we set it
     ROS_INFO("Delay before MCB serial port initialization");
-    ros::Duration(0.5).sleep();
+    ros::Duration(3.0).sleep();
     ROS_INFO("Initialize MCB serial port '%s' for %d baud",
         serial_params.serial_port.c_str(), serial_params.baud_rate);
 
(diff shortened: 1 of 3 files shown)


src/motor_hardware.cc
```diff
@@ -449,8 +449,10 @@ void MotorHardware::setMaxFwdSpeed(int32_t max_speed_fwd) {
 
 // Setup the Wheel Type. Overrides mode in use on hardware  
 // This used to only be standard but THIN_WHEELS were added in Jun 2020
-void MotorHardware::setWheelType(int32_t wheel_type) {
-    ROS_INFO("setting MCB wheel type %d", (int)wheel_type);
+void MotorHardware::setWheelType(int32_t wheel_type, bool showLog) {
+    if (showLog == true) {
+        ROS_INFO("setting MCB wheel type %d", (int)wheel_type);
```
rohbotics: Could we just use ROS_INFO_ONCE here?

src/motor_node.cc
```diff
@@ -81,6 +81,10 @@ int main(int argc, char* argv[]) {
 
     ros::Rate ctrlLoopDelay(node_params.controller_loop_rate);
 
+    // Until we have a holdoff for MCB message overruns we do this delay to be cautious
+    // Twice the period for status reports from MCB
+    float mcbStatusPeriodSec = 0.02;
```
rohbotics: Instead of creating a ros duration every time, this should be a ros::Duration. 

`ros::Duration mcbStatusPeriodSec(0.02);`

Then all the lines below should be `mcbStatusPeriodSec.sleep()`

src/motor_node.cc
```diff
@@ -181,47 +185,55 @@ int main(int argc, char* argv[]) {
...
     }
 
     ros::Time last_time;
     ros::Time current_time;
-    ros::Time last_sys_event_query_time;
     ros::Duration elapsed;
-    ros::Duration elapsed_since_sys_event_query;
-    ros::Duration sysEventQueryPeriod(15.0);     // A periodic query of MCB
+    ros::Time last_sys_maint_time;
+    ros::Duration elapsed_since_last_action;
```
rohbotics: There is no point in declaring this way up here, we should just declare it where we use it.

rohbotics (changes requested):

src/motor_hardware.cc
mjstn: Ok.  I was having mixed feelings that none of the lower level drivers should output a ros log and all should have an override to optionally suppress the log.  General ROS_INFO log for every call is not a decision to be made at the low level. Anyway, for another day perhaps.

src/motor_node.cc
mjstn: Ok, that was silly of me.

src/motor_node.cc
mjstn: I'm mixed on this but did it.

rohbotics (approved):

PR #100, opened 2020-08-08
Fix wheel direction issues

mjstn:
Purpose Of This Pull Request

1) Add support for setting that v38 firmware adds which is to make wheels go in reverse direction.
This is used of the wheels themselves rotate in opposite direction from our standard wheels they can still be used.

2) Add realtime visibility for the key PID parameters into /diagnostics topic and better logs for when they change be in rosout.log

3) Enhance test_motor_board.py script also adds support for wheel direction and a new command of  'O' which will decode option bits, one of which is wheel direction,  and make the tool easier to read the options that show up as bits in HW_OPTIONS register.  A reg dump is also added.


mjstn:
Due to recent events at a customer and this review not ready for other things  I am piggybacking a much needed visibility into PID parameters where we can see current PID parameters in /diagnostics and also there are better, more grepable, logs in ROS log to make all PID parameters visible there as they change (perhaps from dynamic config).    Sorry for making it more complex but this is important for issues going on now with a key customer.

src/motor_hardware.cc
```diff
@@ -780,6 +835,8 @@ void MotorDiagnostics::motor_power_status(DiagnosticStatusWrapper &stat) {
         stat.summary(DiagnosticStatusWrapper::WARN, "Motor power off");
     }
 }
+
+
 // motor_encoder_mode returns 0 for legacy encoders and 1 for 6-state firmware
```
rohbotics: Please update this comment.

src/motor_hardware.cc
```diff
@@ -794,6 +851,10 @@ void MotorDiagnostics::firmware_options_status(DiagnosticStatusWrapper &stat) {
     } else {
         option_descriptions +=  ", Standard wheels";
     }
+    if (firmware_options & MotorMessage::OPT_WHEEL_DIR_REVERSE) {
+        // Only indicate wheel reversal if that has been set as it is non-standard
+        option_descriptions +=  ", Reverse wheels";
```
rohbotics: Can you make it "reverse direction" Otherwise right now we might get "Standard Wheels, Reverse wheels" which is a little confusing.

rohbotics (changes requested):

src/motor_hardware.cc
mjstn: Humm.  Even that too is not enough info.  I see your point.   How about  'Reverse polarity wheels'?
I am trying to say in a short number of words enough to convey this:
'Wheels that when driven by the standard motor controller turn in the reverse way from standard wheels'

src/motor_hardware.cc
mjstn: Very good catch, thanks for finding this.

src/motor_hardware.cc
mjstn: I'll try 'Reverse polarity wheels'

rohbotics (approved):

Issue #59, opened 2019-04-11
No velocity published in Joint_state


davecrawley:
When we try to view velocity in joint_state we only get 0.0 even though the robot is clearly moving at high speed.

mjstn:
At this time joint state position is in radians.
The ROS docs say velocity is in rad/sec.
So I have it working now and it will publish velocity in rad/sec

This actually makes sense because the joint is a rotating wheel not the robot

mjstn:
This is fixed in **branch  fixJointStateVelocities**  and is in the review state at this time

mjstn:
This has been merged and thus this issue is being moved to done now.

mjstn:
This was done long ago and moving to close finally.

PR #101, opened 2020-08-15
Adding wheel joint state velocity to fix ubiquity_motor issue #59

mjstn:
Adding joint state velocity in rad/sec because up to now it has always remained as 0.0

### DIFF
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -141,6 +141,8 @@ public:
     void setOptionSwitchReg(int32_t option_switch);
     void requestSystemEvents();
     void setSystemEvents(int32_t system_events);
+    void getWheelJointPositions(double &leftWheelPosition, double &rightWheelPosition);
+    void setWheelJointVelocities(double leftWheelVelocity, double rightWheelVelocity);
     int firmware_version;
     int firmware_date;
     int firmware_options;
@@ -184,6 +186,12 @@ private:
 
         Joint() : position(0), velocity(0), effort(0), velocity_command(0) {}
     } joints_[2];
+
+    // MessageTypes enum in class to avoid global namespace pollution
+    enum WheelJointNumbers {
+        LEFT_WHEEL_JOINT  = 0,
+        RIGHT_WHEEL_JOINT = 1
+    };
 
     ros::Publisher leftError;
     ros::Publisher rightError;
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -188,9 +188,9 @@ private:
     } joints_[2];
 
     // MessageTypes enum in class to avoid global namespace pollution
-    enum WheelJointNumbers {
-        LEFT_WHEEL_JOINT  = 0,
-        RIGHT_WHEEL_JOINT = 1
+    enum WheelJointLocation {
+        Left  = 0,
+        Right = 1
     };
 
     ros::Publisher leftError;
--- a/src/motor_node.cc
+++ b/src/motor_node.cc
@@ -242,8 +242,8 @@ int main(int argc, char* argv[]) {
     }
 
     float expectedCycleTime = ctrlLoopDelay.expectedCycleTime().toSec();
-    float minCycleTime = 0.75 * expectedCycleTime;
-    float maxCycleTime = 1.25 * expectedCycleTime;
+    ros::Duration minCycleTime = ros::Duration(0.75 * expectedCycleTime);
+    ros::Duration maxCycleTime = ros::Duration(1.25 * expectedCycleTime);
 
     // Clear any commands the robot has at this time
     robot->clearCommands();
@@ -263,7 +263,7 @@ int main(int argc, char* argv[]) {
     // Setup to be able to do periodic operations based on elapsed times
     ros::Time current_time;
     ros::Duration sysMaintPeriod(60.0);     // A periodic MCB maintenance operation
-    ros::Duration jointUpdatePeriod(0.5);   // A periodic time to update joint velocity
+    ros::Duration jointUpdatePeriod(0.25);  // A periodic time to update joint velocity
 
     ros::Time last_loop_time = ros::Time::now();
     ros::Duration elapsed_loop_time;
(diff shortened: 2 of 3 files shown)


src/motor_node.cc
```diff
@@ -249,25 +249,42 @@ int main(int argc, char* argv[]) {
...
     while (ros::ok()) {
         current_time = ros::Time::now();
-        elapsed = current_time - last_time;
-        last_time = current_time;
-        robot->readInputs();
+        elapsed = current_time - last_loop_time;
+        last_loop_time = current_time;
         float elapsedSecs = elapsed.toSec();
+
+        // Determine and set measured wheel velocities in rad/sec from hardware positions in radians
```
rohbotics: This block of code should be inside readInputs, just like all the other joint states stuff.

src/motor_node.cc
```diff
@@ -249,25 +249,42 @@ int main(int argc, char* argv[]) {
...
     // Clear any commands the robot has at this time
     robot->clearCommands();
 
-    last_time = ros::Time::now();
-    ros::Time last_sys_maint_time = last_time;
+    last_loop_time = ros::Time::now();
+    ros::Time last_sys_maint_time = last_loop_time;
     ros::Duration elapsed_since_last_action;
 
+    double leftWheelVelocity  = 0.0;
```
rohbotics: I don't like declaring variables far from where they are used, it gives them more scope than they really need, and makes their use less clear. These should be declared where they are used (inside the loop).

include/ubiquity_motor/motor_hardware.h
```diff
@@ -185,6 +187,12 @@ class MotorHardware : public hardware_interface::RobotHW {
         Joint() : position(0), velocity(0), effort(0), velocity_command(0) {}
     } joints_[2];
 
+    // MessageTypes enum in class to avoid global namespace pollution
+    enum WheelJointNumbers {
```
rohbotics: I think this is a good idea. 

Maybe we use a enum class named `WheelJoint` and then you can refer to the constants with WheelJoint::Left and WheelJoint::Right so that the source of the constant is clear.

rohbotics (changes requested):

src/motor_node.cc
mjstn: I did not put it in readInputs because it requires history of elapsedSecs.

After seeing this value jump I am now thinking I want even greater time spacing due to our crappy encoder counts.
By the way encoder ticks are getting even worse with non-geared avalon we go to a whopping 86 ticks/rev. Good lord.

So with your comment and seeing the velocity jump a bit even now I am thinking of spacing out even more the measurement to 250ms or even 500ms.  It will be more accurate at low speeds and we have zero hope in hell of having it be of much value at all if it is calculated very ODOM report from the MCB.

So it could be in ODOM message in readInputs() or in every readInputs() but I do insist it requires a timer and I am not real happy about that timer being in MotorHardware as it's hard to support being dependent on ODOM report rate so if MCB fails velocity could get stuck at some high value.

src/motor_node.cc
mjstn: In this case, religion trades off with code readability.  
All the other vars for this are there so there is an argument to keep them together and not bulk up the while loop.
I will yield to your request but do not agree.

include/ubiquity_motor/motor_hardware.h
mjstn: Like this sort of naming but think just WheelJoint is not descriptive enough maybe Ill use WheelJointLocation  because it is then even more readable.   thanks, good input.

src/motor_node.cc
mjstn: Have pushed changes so that updates to 250msec and it is much more stable so that fixes earlier where it was updated too recently and encoder resolution was making the joint velocity have significant jitter.

src/motor_node.cc
mjstn: This is now pushed for at least the ones that can move

src/motor_node.cc
```diff
@@ -249,31 +245,59 @@ int main(int argc, char* argv[]) {
...
+            leftLastWheelPos  = leftWheelPos;
+            rightLastWheelPos = rightWheelPos;
+        }
+
         robot->readInputs();
-        float elapsedSecs = elapsed.toSec();
-        if (minCycleTime < elapsedSecs && elapsedSecs < maxCycleTime) {
-            cm.update(current_time, elapsed);
+        if ((minCycleTime < elapsed_loop_time.toSec()) && 
+            (elapsed_loop_time.toSec() < maxCycleTime)) {
```
rohbotics: minCycleTime and maxCycle time should be ros::Durations, that way we don't have to do toSec for the comparision.

rohbotics (changes requested):

src/motor_node.cc
mjstn: I have done this because 'you are a nice guy'.

MoffKalast: Lmao

mjstn: He is a nice guy.  If not I would frankly have said:   'Really?'  and ignored this so we could progress. We loose 1000x as many cpu cycles from some silly issue in fiducial recognition than this conversion.

mjstn: Also: This does not require you as well Vid, I ment to take you off.  I only added you cause rohan took 9 days due to some situations going on in his work or non-work life to get to this rather non significant bug that has been on the books for a hella long time anyway.

mjstn: Crud.  I cannot figure out how to remove you from this review Vid.   i'll try again tomorrow, ignore this one or just approve, your call.

MoffKalast (approved):

rohbotics (approved):
Sorry for the last point, yes it was a nit. I should have specified that it was non-blocking.

In general I find minor code cleanup points to be pretty useful in code review. In this case my comment had nothing to do with speed of execution or anything like that. It is purely because I have seen plenty of bugs caused by mixing floats in with time comparisons, and although those are unlikely to show up in here it is much easier to just use ros::Durations everywhere to avoid future issues (or std::chrono or whatever your time lib of choice is).

mjstn:
Thanks for the explanation Rohan.  I see your point now as not being about code efficiency.  Thanks for your inputs.

PR #102, opened 2020-08-26
Reinit mcb on mcb reset

mjstn:
We had before detected that the MCB had been reset using a system event.  We only stated it in the log.

This change when seeing an MCB reset re-initializes the MCB with any ROS parameters that the host would normally have only setup on the MCB at startup.  Now a large routine called initMcbParameters() is used.

The next involved change is a lot of variable renames mostly to indicate they are globals in many case.
This change will mean there will be merge conflicts.

before the push of this branch did this merge:  git merge --no-ff fixJointStateVelocities


rohbotics:
Mark you will have to rebase this PR. `git fetch` `git rebase origin/kinetic-devel` it will complain about conflicts, you will have to manually resolve those.

PR #103, opened 2020-08-29
Reset and reinitialize MCB if we see it has itself been reset

mjstn:
Form initMcbParameters() to be used in place of inline code at startup and also to be used when a MCB reset system event is seen from the MCB.      

initMcbParameters() has the task of making sure the MCB is setup with any parameters that the host sets like Pid and many other parameters

Also modified the periodic message to log so it now also shows battery voltage which is so very critical to debug any number of customer complatints.

### DIFF
--- a/src/motor_node.cc
+++ b/src/motor_node.cc
@@ -41,9 +41,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 static const double BILLION = 1000000000.0;
 
-static FirmwareParams firmware_params;
-static CommsParams serial_params;
-static NodeParams node_params;
+static FirmwareParams g_firmware_params;
+static CommsParams    g_serial_params;
+static NodeParams     g_node_params;
+
+// Until we have a holdoff for MCB message overruns we do this delay to be cautious
+// Twice the period for status reports from MCB
+ros::Duration mcbStatusPeriodSec(0.02);
 
 // Dynamic reconfiguration callback for setting ROS parameters dynamically
 void PID_update_callback(const ubiquity_motor::PIDConfig& config,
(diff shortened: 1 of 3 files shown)


src/motor_node.cc
```diff
@@ -53,49 +57,146 @@ void PID_update_callback(const ubiquity_motor::PIDConfig& config,
...
 int main(int argc, char* argv[]) {
     ros::init(argc, argv, "motor_node");
     ros::NodeHandle nh;
 
-    firmware_params = FirmwareParams(nh);
-    serial_params = CommsParams(nh);
-    node_params = NodeParams(nh);
-
-    ros::Rate ctrlLoopDelay(node_params.controller_loop_rate);
+    g_firmware_params = FirmwareParams(nh);
```
rohbotics, line 183: Nit: these aren't really globals, so not sure the `g_` prefix makes sense.

src/motor_node.cc
```diff
@@ -53,49 +57,146 @@ void PID_update_callback(const ubiquity_motor::PIDConfig& config,
...
     } else {
         ROS_ERROR("Unsupported dynamic_reconfigure level %u", level);
     }
 }
 
+//  initMcbParameters()
+//
+//  Setup MCB parameters that are from host level options or settings
+//
+void initMcbParameters(std::unique_ptr<MotorHardware> &robot )
```
rohbotics, line 82: I am thinking that this might be better to put in MotorHardware directly, there is no need to clutter the node stuff with it.

Also if you keep it out here, there is no need to pass a unique_ptr by ref, you should be passing a MotorHardware by ref into this function.

`void initMcbParameters(MotorHardware& robot)`

Which can be called in main by `initMcbParameters(*robot)`

rohbotics (approved):

src/motor_node.cc
mjstn: Ok, got carried away.  I like to use g_ when something is a var in process global space and it should not be g_ if it is a member variable of any class.

src/motor_node.cc
mjstn: Well here is my take on why it belongs in motor_node.cc:
I see motor_node.cc the 'mother function' and in fact the node itself.    This is where the ROS parameters should be recognized.  I see motor_hardware.cc as the 'bunch of hardware specific worker routines.
Because it is in a function now I do not see this as 'cluttering'   in fact it was more cluttered before with a lot of this right in the main for the node.    So I see motor_node.cc as the place where high level node parameters are passed to the node and make their way to lower level.   I also see motor_node.cc as the place where decisions that vary the node operation should mostly happen.   It is  'the app' is my thought.  All else (hardware, messages, serial) are the things that are envolked by the main 'app'.     I hope you agree.

src/motor_node.cc
mjstn: Actually I just looked.   These were globals before without g_  so I added g_
firmware_params, serial_params and node_params were all globals before but with no g_
the g_ makes it clear now that we are talking about the static  FirmwareParams in motor_node.cc

They have been globals or at least   static casts in motor_node.cc (a scoped global if you will) since 2017

mjstn: I am loosely using global here to mean not in a class.  Not totally open globals but static cast and for the sake of the code in motor_node.cc for those purposes they are global.

PR #104, opened 2020-08-31
I2C Lock

dorkamotorka:
Merged the write() and read() calls. Changed the chipRegAddr argument to the function, so that it represents the i2c slave's register from which we want to read.

### DIFF
--- a/src/motor_hardware.cc
+++ b/src/motor_hardware.cc
@@ -871,6 +871,8 @@ static int i2c_BufferRead(const char *i2cDevFile, uint8_t i2c8bitAddr,
 {
     int fd;                                         // File descriptor
     int retCode = 0;
+    int slaveAddress = i2c8bitAddr >> 1;              // Address of the I2C device
+
     if ((fd = open(i2cDevFile, O_RDWR)) < 0) {      // Open port for reading and writing
       ROS_ERROR("Cannot open I2C def of %s with error %s", i2cDevFile, strerror(errno));
       retCode = -1;
@@ -879,10 +881,8 @@ static int i2c_BufferRead(const char *i2cDevFile, uint8_t i2c8bitAddr,
 
     uint8_t buf[1];                                   // Buffer for data being written to the i2c device
     uint8_t outbuf[1];
-    int slaveAddress = i2c8bitAddr >> 1;              // Address of the I2C device
     struct i2c_msg msgs[2];
     struct i2c_rdwr_ioctl_data msgset[1];
-
     msgs[0].addr = slaveAddress; // slave adress
     msgs[0].flags = 0; // write bit
     msgs[0].len = 1; // number of data bytes written to I2C slave address 
--- a/src/motor_hardware.cc
+++ b/src/motor_hardware.cc
@@ -871,7 +871,7 @@ static int i2c_BufferRead(const char *i2cDevFile, uint8_t i2c8bitAddr,
 {
     int fd;                                         // File descriptor
     int retCode = 0;
-    int slaveAddress = i2c8bitAddr >> 1;              // Address of the I2C device
+    int slaveAddress = i2c8bitAddr >> 1;            // Address of the I2C device
 
     if ((fd = open(i2cDevFile, O_RDWR)) < 0) {      // Open port for reading and writing
       ROS_ERROR("Cannot open I2C def of %s with error %s", i2cDevFile, strerror(errno));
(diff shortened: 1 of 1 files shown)


mjstn (commented):
In the meeting did you say this has never been actually run on the MCB?    That is important.  Hold off on any merge till I actually run this if you have no ability to do so.   The lock mode must apply to reading of the 8-bit IO expander as well as the OLED display or locking does not buy us much.  

What I had envisioned for I2C is that we can do what you have done below but must also do it in oled_display_node repository.

A longer term project I would like to see is that all the I2C interface code is in a library that can be used by the current 2 user nodes as well as possible future  accelerometer/gyro code what will soon exist.  

The accel/gyro code will need to do multi-byte reads so I'm not sure why  i2c_BufferRead no longer has a byte count but I do see the ioctl has ability to specifify multi-byte reads.

Anyway, a lot to think about and I will do more of a review and test soon.

Thanks,
Mark

rohbotics:
Mark, this already buys us a bunch even without the oled node doing it. It's not a lock but an actual kernel level transactional interface. This means that nothing else will be allowed down the bus until this read/write cycle is complete. It's not just adding a lock that other users must make sure to grab. 

I don't see much use in spinning out such a small piece of code into a library, I am ok with keeping it in each node that needs to do I2C. Because ubiquity_motor specifically doesn't need multi byte reads, I think its worth simplifying the code in here to remove that support.

Rohan

mjstn:
I should have been a bit more clear.   Yes this is of value and I will test and we can approve and push at that time, likely in a day or two. 

I was bringing up that at some later date in a different change be used in oled module.   NOT for this, for later.

The other thing I discussed is a larger project for another time, not for this change.   I was sort of 'thinking out loud' about future work that could be done to form a library usable by at least 3 of our repositories, one of which does not exist yet (accelerometer/gyro).      Pay no attention to that I have to sort out starting a backlog task would be best.

src/motor_hardware.cc
```diff
@@ -63,7 +64,7 @@ int32_t  g_odomEvent = 0;
 // This utility opens and reads 1 or more bytes from a device on an I2C bus
 // This method was taken on it's own from a big I2C class we may choose to use later
 static int i2c_BufferRead(const char *i2cDevFile, uint8_t i2cAddr, 
-                          uint8_t* pBuffer, int16_t chipRegAddr, uint16_t NumByteToRead);
+                          uint8_t* pBuffer, int16_t chipRegAddr);
```
mjstn, line 68: Why are we loosing the ability to read more than one byte?   Is that a restriction of this method of reading?

src/motor_hardware.cc
```diff
@@ -863,53 +864,52 @@ void MotorDiagnostics::firmware_options_status(DiagnosticStatusWrapper &stat) {
...
-      goto exitWithFileClose;
-    }
+    uint8_t buf[1];                                 // Buffer to be written to i2c device 
+    uint8_t outbuf[1];                              // Buffer to be read into from i2c device
+    struct i2c_msg msgs[2];
+    struct i2c_rdwr_ioctl_data msgset[1];
+    msgs[0].addr = slaveAddress;                    // Slave adress
+    msgs[0].flags = 0;                              // Write bit
+    msgs[0].len = 1;                                // Number of data bytes written to I2C slave address 
+    msgs[0].buf = buf;                              // Writting to register
```
mjstn: Does this ioctl  and  msgs[0].len = 1 write to the device ?
If so that can cause a problem.  Writing to this device configures pins to be inputs or outputs.
We should not write to this  device over I2C.

src/motor_hardware.cc
dorkamotorka: Well previously we had two methods write() and read() in order to read from the chips register. And it was a must to check whether the number of bytes we requested then actually received, because of race condition. And it was straight forward to check it, because the read() method returned the number of bytes read. Now there is only one call ioctl() that combines both of the methods(write() and read()) and locks any other device that is trying to interfere in the meantime. Since ioctl() is the only call in this case, if the call was successful it returns a non-negative number, while if the call is unsuccessful it returns the -1. From this point, there is no straight-forward method to check how many bytes were read, but it think its up to the testing to confirm the reliabilty of this method.

src/motor_hardware.cc
dorkamotorka: ioctl() in this case with macro I2C_RDWR first writes to the slave device register and then reads from it.
msgs[0].len = 1 is the length of the message written to the device in bytes. So in our case, we are writting one byte to the i2c slave device register.

That's how the I2C protocol is defined. In order to read from the i2c slave device register, you must first tell it from which of its internal addresses you want to read, which is done by writing to it. I think there is no way to actually immediatelly read from the slave device. The only problem I see it might occur is clock streching(which might be automatic on our i2c devices and the problem is gone in that case).

mjstn: The reason I ask is that the PCF8574 is a very simple device and does not have internal chip addressing for registers.  There is only one register to write and that confures the pins as inputs or outputs.  We do not want to write to this chip except to configure it for all inputs in this case.  Find TI datasheet for PCF8574 and you will find this usage info:

This device does not have internal configuration or
status registers. Instead, read or write to the device I/Os directly after sending the device address

That is why the interface to this read routine had a special trap for chipRegAddr so it could suppress the writing of the 'register address' to the chip which in this case would re-configure the chip to all outputs if we wrote out a 0x00.  If we write to this chip at all it should be 0xff for all inputs BUT we do not want to write to it at all or general usage of the routine could not deal with a PCF8574 we use in a different way.   I have 3 pins on this chip that could be outputs because they go to a jack so the routine to access the part must not make assumptions on chip configuration.

src/motor_hardware.cc
mjstn: Yes I get the reason for the ioctl.  I once wrote a complex fibre channel driver for a unix like system that had it's reads and writes accessed using an IOCTL.   So I get that and I am glad we are going to use this interface.   If this routine is changed to only read one byte it is ok for this usage but then change it's name to  indicate it only reads one byte and change assorted comments in the routine that discuss number of bytes such as  'reads 1 or more bytes from a device on an I2C bus'.  We will at a later time develop a general purpose read call but it's ok to just read one byte for this change.   I edited this comment because I talked to Rohan and he is ok with just having it read one byte.

mjstn:
There was a bug in the earlier routine that misled you to feel the write was needed.   That is 'my bad'.  
The comment was what we want and the code itself should have been a test to only do the write if chipRegAddr >= 0.   

Here is buggy code that you were apparently simply making your code do the same thing.   Again, my bad.
if (chipRegAddr < 0) {     // Suppress reg address if negative value was used

dorkamotorka:
So now there is only a read call (probably I could make a bit more informative) - the same as what the last version of Mark's code should do (without the above mentioned bug). I think it is up to the testing to show whether this works. 

If the code doesn't work, we should also try to change the MACRO in https://github.com/UbiquityRobotics/ubiquity_motor/blob/b7edbdbcec08d89b3702b68b3a3bbf27a6e51dab/src/motor_hardware.cc#L877 from O_RDWR (open for writing and reading) to O_RDONLY (open for reading only).

I do have a question.. Is our hardware compatible with SMBus protocol or is it I2C only?

mjstn:
The part that is being discussed,  PCF8574, is a very simple minded pure I2C device.
I will update for this branch and test the 'read only' mode.
One thing that is interesting in the old bug is that the reason it worked is a  -1 was sent which is hex 0xFF so in fact the write that should not have been done was ok in that it told the part all bits were inputs (1)   Funny in that way.

mjstn (approved):
This functions properly over I2C for the mode we use it for at this time.
We will form flexible and more general purpose read and write routines using the Ioctl in OLED repository.

I don't think you need to make Rohan review this.  Up to you to decide.

rohbotics (approved):
:+1:

src/motor_hardware.cc
```diff
@@ -863,53 +864,42 @@ void MotorDiagnostics::firmware_options_status(DiagnosticStatusWrapper &stat) {
...
-
-    int fd;                                         // File descrition
-    int  address   = i2c8bitAddr >> 1;              // Address of the I2C device
-    uint8_t buf[8];                                 // Buffer for data being written to the i2c device
+    int fd;                                         // File descriptor
+    int retCode = 0;
+    int byteRead = 0;
+    int slaveAddress = i2c8bitAddr >> 1;            // Address of the I2C device
 
     if ((fd = open(i2cDevFile, O_RDWR)) < 0) {      // Open port for reading and writing
```
dorkamotorka, line 877: Can you please also test this line 877 like that: change O_RDWR into O_RDONLY 
Because:
- O_RDWR is opening file for reading and writing
- O_RDONLY is opening file for reading

mjstn: I can test to verify it works in the only mode we run it in now which is read one byte.
BUT you seem to have not changed it so there would be no value to retest.   
We should check the same thing that would be merged later and not let me mess with it on my own code copy.

PR #105, opened 2020-09-01
Motor current publishing

mjstn:
Add a message that given new v38 firmware will publish to topic /motor_state a message with motor currents, motor pwm settings and motor rotational velocities.


include/ubiquity_motor/motor_hardware.h
```diff
@@ -86,6 +88,14 @@ struct MotorDiagnostics {
     float battery_voltage_low_level = 22.5;
     float battery_voltage_critical = 21.0;
 
+    // Wheel current states
+    double motorCurrentLeft  = 0.0;
+    double motorCurrentRight = 0.0;
+    double motorAmpsPerAdcCount = 0.0238;   // 0.1V/Amp  2.44V=1024 count so 41.97 cnt/amp
```
rohbotics: (optional) These are constants, they probably shouldn't be in this struct. If you chose to ignore this for now, please put a TODO comment here saying that it should be done.

src/motor_hardware.cc
```diff
@@ -173,6 +177,24 @@ void MotorHardware::setWheelJointVelocities(double leftWheelVelocity, double rig
     return;
 }
 
+// Publish motor state conditions
+void MotorHardware::publishMotorState(void) {
+    ubiquity_motor::MotorState mstateMsg;
+
+    mstateMsg.header.frame_id = "motor_state";
+    mstateMsg.header.stamp    = ros::Time::now();
+    mstateMsg.header.seq      = 0;
```
rohbotics: Don't put seq, its just confusing because ros will overwrite it anyway.

src/motor_hardware.cc
```diff
@@ -173,6 +177,24 @@ void MotorHardware::setWheelJointVelocities(double leftWheelVelocity, double rig
     return;
 }
 
+// Publish motor state conditions
+void MotorHardware::publishMotorState(void) {
+    ubiquity_motor::MotorState mstateMsg;
+
+    mstateMsg.header.frame_id = "motor_state";
```
rohbotics: If you really want it to have a frame make it "base_link" but otherwise the data in the message is frameless so we should have an empty string.

rohbotics (changes requested):
Mostly pretty small and quick things, you are good to merge after those are fixed.

include/ubiquity_motor/motor_hardware.h
mjstn: Half ok.   The amps per adc count will not change.
The zero offset is something I am seeing can be off.  Rather than work out a full calibration I wanted it to be able to be different.  We may do that cal if currents near zero are important enough to require.
So I will make one a constant but the zero offset will remain in the class.

src/motor_hardware.cc
mjstn: Ok.  While testing after i pushed I saw this and was going to change it anyway.

src/motor_hardware.cc
mjstn: Ok.  Silly me.   This is the same sort of frame_id used for fiducials in that it is a spacial frame.  I'll make it empty.
I was mostly focused on timestamp and just put this in there not thinking much about what it would be used for later.

mjstn:
You never clicked approve so I await that 'in the future'

rohbotics (approved):
:+1:

PR #106, opened 2020-09-03
Added the motor positon to MotorState

rohbotics:
All the information we have about the motors is now published in MotorState.


mjstn (approved):
Ok.   I made it float32 at first because there is frankly nowhere near 64bit accuracy and I hate wasting bits and ethernet bandwidth (these topics go across to other pi) but I guess these days that should not be a consideration.   So I am ok with float64

PR #109, opened 2020-09-22
Open device for read-only.

### LINKED ISSUE #108: Test I2C Communication with O_RDONLY
Macro O_RDWR should be changed to O_RDONLY and tested, because we are only doing the read() call afterwards-
    O_RDWR is opening file for reading and writing
    O_RDONLY is opening file for reading

https://github.com/UbiquityRobotics/ubiquity_motor/blob/437d31cdb6ff262c01afd4769900ee93d08b0ea1/src/motor_hardware.cc#L897

dorkamotorka:
It adds a small change to the way file/device is open(). O_RDWR enabled to write and read from the device, whereas O_RDONLY just enables reading. Mark already tested it.


mjstn (approved):
This works with read only access per some tests I had done.

PR #110, opened 2020-10-12
Wheel overheating Level 1 fix

mjstn:
This is a fix where if the robot is sitting at zero velocity for a few seconds we will try to avoid wheel overheat by setting the prior wheel encoder position to the current wheel encoder position within the MCB.    

This 4 wheel drive design by nature of fixed axles and wheels under load not slipping easily at all is prone to significant currents being required in a static state where wheels fight for their own position.  

The code for this needs to be in host because the host knows the big picture.    Dangerous to have this automatically in the MCB.

Mcb firmware from 20201006 or later is required so that it will accept a new MCB command to do the 'null' 

This fix is guarenteed to cause some level of odom 'slip' but to be frank, in this 4 wheel design odom is a joke for any rotations thus we don't give up much of anything in this respect.


include/ubiquity_motor/motor_hardware.h
```diff
@@ -201,6 +204,12 @@ class MotorHardware : public hardware_interface::RobotHW {
         Joint() : position(0), velocity(0), effort(0), velocity_command(0) {}
     } joints_[2];
 
+    // MessageTypes enum for refering to motor or wheel number
+    enum MotorOrWheelNumber {
+        Motor_M1 = 1,
```
MoffKalast, line 209: Wouldn't this perhaps make more sense as Motor_L and Motor_R?

src/motor_hardware.cc
```diff
@@ -457,6 +457,24 @@ void MotorHardware::writeSpeeds() {
...
 }
 
+// areWheelSpeedsZero()  Determine if all wheel joint speeds are below given threshold
+//
+int MotorHardware::areWheelSpeedsLower(double wheelSpeedRadPerSec) {
+    int retCode = 0;
+
+    // This call pulls in speeds from the joints array maintained by other layers
+
+    double  left_radians  = joints_[WheelJointLocation::Left].velocity_command;
```
MoffKalast, line 467: Double spaces after 'double'? Ironic.

include/ubiquity_motor/motor_parameters.h
```diff
@@ -175,27 +175,31 @@ struct NodeParams {
...
-
+    std::string drive_type;
     std::string left_joint;
     std::string right_joint;
 
     NodeParams()
         : controller_loop_rate(10.0),
           wheel_type("firmware_default"),
           wheel_direction("firmware_default"),
+          drive_type("2wd"),
```
MoffKalast, line 186: So shouldn't this be set to 4wd here? I mean by default keep it 2wd sure, but some config file needs to change it to 4wd when launching and I'm not certain we do that in the current core/base.launch in avalon_bringup. Do we need to make an addition?

include/ubiquity_motor/motor_parameters.h
```diff
@@ -175,27 +175,31 @@ struct NodeParams {
...
 	  left_joint("left_wheel_joint"),
 	  right_joint("right_wheel_joint"){};
 
     NodeParams(ros::NodeHandle nh) : controller_loop_rate(10.0),
         wheel_type("firmware_default"),
-        wheel_direction("firmware_default") {
+        wheel_direction("firmware_default"),
+        drive_type("2wd") {
```
MoffKalast, line 193: Same as above.

src/motor_node.cc
```diff
@@ -203,6 +204,21 @@ int main(int argc, char* argv[]) {
...
+    // This leads to motor overheating.  This code below sacrifices accurate odometry which
+    // in not achievable in such robots anyway to relieve high wattage static drive currents
+    // at zero velocity that are due to inability of wheels to slip tiny amounts.
+    int32_t wheel_slip_nulling = 0;
+    if ((robot->firmware_version >= MIN_FW_WHEEL_NULL_ERROR) && (node_params.drive_type == "4wd")) {
+        wheel_slip_nulling = 1;
+        ROS_INFO_ONCE("Wheel slip nulling will be enabled for this 4wd system when velocity remains at zero.");
+    }
+
+    wheel_slip_nulling = 1;   // !!! DEBUG HACK !!! Force to 1 for avalon code till param flows to here
```
MoffKalast, line 218: Ah, I see the fix for that 2wd param.

MoffKalast (approved):
Ok in general we can merge, but I'd just like to get on the same page about where we need to make the param changes. I assume an additional param in [the motor node section of core.launch](https://github.com/UbiquityRobotics/avalon/blob/master/avalon_bringup/launch/core.launch) for both rear and front?

include/ubiquity_motor/motor_hardware.h
mjstn: Yes it would be the M1 and M2 are greatly ingrained in the code and it would be very messy to try to change it.   The back story here is there are numerous screw ups in right and left wheel that get sorted out at this time but are 'dirty'.   So comments in some places may refer to M1 but if we change it to L or R those comments would be disturbing.

src/motor_hardware.cc
mjstn: Yes.  Deal with it.

include/ubiquity_motor/motor_parameters.h
mjstn: This may be useful in Avalon branch with a comment saying in master we need to keep it 2wd.   'Someday' we want to merge avalon back to main and we have to be careful to explain all the oddities.

PR #111, opened 2020-12-05
Fix wheel slip overheat1

mjstn:
This is badly needed new high torque wheel modes for key customer.  
by default this change does not go to the MCB PID control register because base.yaml does not have definition for those bits for main customers.   The key customer pid_control  value s in base.yaml should be 512 to activate this best mode so far.


mjstn:
A review is requested but before this can merge to main for magni platform we need to have test group test this more on a real magni to verify compatibility.  Many changes here were done for a 4WD robot and we must be sure these still work fine on a standard Magni product base.

mjstn:
WARNING!   This totally blows up on a magni!!!!!     DO NOT MERGE THIS!!!!!    Just do review and mention any findings.   Thanks.    I will point out that current top of tree master does work GREAT with v39.   I think I must pull in changes to current top of master tree then ask for review.

include/ubiquity_motor/motor_parameters.h
```diff
@@ -175,18 +180,35 @@ struct NodeParams {
...
+          drive_type("2wd"),
+	  left_joint("left_wheel_joint"),
+	  right_joint("right_wheel_joint"){};
 
-    NodeParams() : controller_loop_rate(10.0), wheel_type("firmware_default"), wheel_direction("firmware_default"){};
     NodeParams(ros::NodeHandle nh) : controller_loop_rate(10.0),
         wheel_type("firmware_default"),
-        wheel_direction("firmware_default") {
+        wheel_direction("firmware_default"),
+        drive_type("2wd") {
```
MoffKalast, line 198: Assuming this is set to 4wd in a launch file?

src/motor_hardware.cc
```diff
@@ -102,8 +102,12 @@ MotorHardware::MotorHardware(ros::NodeHandle nh, CommsParams serial_params,
     battery_state = nh.advertise<sensor_msgs::BatteryState>("battery_state", 1);
     motor_power_active = nh.advertise<std_msgs::Bool>("motor_power_active", 1);
 
+    motor_state = nh.advertise<ubiquity_motor::MotorState>("motor_state", 1);
+    leftCurrent = nh.advertise<std_msgs::Float32>("left_current", 1);
```
MoffKalast, line 106: Doesn't this duplicate the motor_state topic data?

src/motor_hardware.cc
```diff
@@ -690,13 +796,31 @@ void MotorHardware::sendParams() {
         commands.push_back(maxpwm);
     }
 
+    if (cycle == 7 &&
```
MoffKalast, line 799: Why 7 though? I thought 42 was the answer.

MoffKalast (commented):
A few nitpicks, but also this should be merged into 'avalon', not 'kientic-devel' if merged. Also we've been using this branch on the polaris images for a while now and it definitely works fine there, so as long as we keep it polaris only we're fine.

PR #115, opened 2021-01-11
Add system control for mcb disable

mjstn:
We add ability to disable and re-enable motor node ownership of MCB through serial port close and re-open.
All control is done using motor_node listing to topic  /system_control   The following strings control operation
-  motor_control  disable
- motor_control   enable

### DIFF
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -118,7 +118,7 @@ public:
                   FirmwareParams firmware_params);
     virtual ~MotorHardware();
     void closePort();
-    void openPort();
+    bool openPort();
     void clearCommands();
     void readInputs();
     void writeSpeeds();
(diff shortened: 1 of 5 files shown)


MoffKalast (commented):
Well I don't see anything out of the ordinary in the code, aside that some parts seem unrelated to the main change (I just hope there won't be merge difficulties to the wheelslipoverheat1 branch).

I don't have a robot to test myself but I'll ask if @JanezCim has some time when he'll in TP these days to at least give it a short whirl on a regular magni before we mainline it.

I'll make sure the update service supports sending the message to the topic in the meantime.

JanezCim:
I will be able to test in the TP tommorow

JanezCim (approved):
Tested this on a magni in tp

When publishing disable with `rostopic pub /system_control std_msgs/String "data: 'motor_control disable'"` the robot did not move on teleop commands and after enabling again with  `rostopic pub /system_control std_msgs/String "data: 'motor_control enable'"` the robot could move again. This did not fail after 3 tries.

PR #116, opened 2021-01-13
Updating fixWheelSlipOverheat1

MoffKalast:
Merging new features into avalon's currently used branch.


MoffKalast:
Okay so I figured it would be possible to just merge and I did solve some of the conflicts, but for a few I can't exactly tell what's right since both sides have conflicting changes. Could you take a look @mjstn?

include/ubiquity_motor/motor_hardware.h
```diff
@@ -125,6 +127,8 @@ class MotorHardware : public hardware_interface::RobotHW {
     void requestFirmwareDate();
     void setParams(FirmwareParams firmware_params);
     void sendParams();
+    void forcePidParamUpdates();
```
mjstn, line 130: These 2 cannot be here unless the source code from motor_hardware.cc that contains these 2 members is also merged.

MoffKalast:
Alright so we'll be needing parts of this on the next image if we want to kill the motor. Seems like there are no remaining conflicts so I'll merge for now.

MoffKalast:
Ah no wait, there ARE still conflicts. Well in that case, probably best not to.

mjstn (commented):
Now on Feb 6, 2021 I am going to form a new branch with a bunch of merges in from Magni features.
This new branch will be called    4wdChassisBranch and thus will quite nicely explain it's purpose.
The branch will be our 4WD branch of Ubiquity Motor until such a time we can merge 4WD control architecture into mainline kinetic-devel codeline and end this insanity of support across kinetic-devel to the 4WD code.

PR #122, opened 2021-02-16
Adding system_control command of speed_control  and cleanup of mcb disable

mjstn:
Adding to topic system_control ability to halt the robot for collision usage from outside this node.

Moving the state for both speed control AND earlier added motor control to disable mcb control by motor node both into NodeParams class.    Also move the defines for topic name and both commands into motor_parameters.h which is better place since now mcbSpeedControl and mcbMotorControl are now members of NodeParams which is also defined in motor_parameters.h

### DIFF
--- a/include/ubiquity_motor/motor_parameters.h
+++ b/include/ubiquity_motor/motor_parameters.h
@@ -33,6 +33,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <ros/ros.h>
 
+// These defines are for a high level topic for control of the motor node at a system level
+#define ROS_TOPIC_SYSTEM_CONTROL  "system_control"    // A topic for system level control commands
+#define MOTOR_CONTROL_CMD         "motor_control"     // A mnumonic for a motor control system command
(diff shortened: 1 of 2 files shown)


MoffKalast (approved):
Seems good, so for future reference `speed_control enable/disable` posted on `/system_control` should be used for pausing the motor controller.

The term speed_control does sound a bit off to me, it's kind of like we're giving up control and letting the motor run by itself. I would've called it something like `drive enable/disable` but it's probably not worth changing.

mjstn:
I agree that it is hard to get a name that explains what is going to happen nicely.
What this thing does (really) is continuously send speed Zero instead of any non-zero value.
My 'reasoning' for this mnumonic was that this keyword of 'speed control' means  we either:
enable  - This allows the motors speed to be set or controlled from cmd_vel
disable -  We override speed and force speed to zero.
To be very explicit we could call it  'zero_speed_override'  but then it is sort of backwards in that we must have this always disabled normally then a user sends  'zero_speed_override  enable'  which then zeros speed.  So that too seems not a good name.

So I am going to leave it like it is now and we discuss it in documents as to what it does.
I have already told the guy he could try it and told him how to do it so it is partly set in still un-hardened cement.

No good answer, I'll leave it.  Thank you for your inputs.

PR #134, opened 2021-06-01
Migration to Python3.8

dorkamotorka:
Migration of python files for Noetic distribution. 
Before merging, we should test on the real robot.


scripts/upgrade_firmware.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/python
+#! /usr/bin/env python
```
rohbotics: Should be `#!/usr/bin/env python3`

scripts/test_motor_board.py
```diff
@@ -38,7 +38,7 @@
 import time
 import serial
 import string
-import thread
+import _thread
```
rohbotics: We should migrate to the new threading module instead of using the Python2 compat mode.

scripts/test_pi_gpio.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
```
rohbotics: #!/usr/bin/env python3

scripts/test_motor_board.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/pythmn -u
+#!/usr/bin/env python
```
rohbotics: #!/usr/bin/env python3

rohbotics (changes requested):

scripts/test_pi_gpio.py
dorkamotorka: Make sense yaa, "scratch scratch"

scripts/upgrade_firmware.py
```diff
@@ -228,22 +228,22 @@ def read_num(self, bytes):
     def send(self):
         self.is_sent = True
         init_bytes = [chr(0x1), self.cmd, 
-                      chr((len(self.out) / 001) % 256),
-                      chr((len(self.out) / 256) % 256)]
+                      chr((len(self.out) / 0o001) % 256),
+                      chr((len(self.out) / 0o256) % 256)]
```
rohbotics: I don't think these numbers should actually be octal. (1 is going to the same in either base, and the 256 didn't have a leading 0 in the original).

scripts/upgrade_firmware.py
```diff
@@ -228,22 +228,22 @@ def read_num(self, bytes):
...
+        if DEBUG: print("Send packet: " + cstr(self.out))
 
         ib_buf = ""
         for c in init_bytes: ib_buf += c
-        if DEBUG: print "Packet without end and checksum is", cstr(ib_buf)
+        if DEBUG: print("Packet without end and checksum is", cstr(ib_buf))
 
         checksum = compute_checksum(ib_buf)
-        if DEBUG: print "Checksum is", checksum, "LSB", (checksum / 001) % 256, "MSB", (checksum / 256) % 256
+        if DEBUG: print("Checksum is", checksum, "LSB", (checksum / 0o001) % 256, "MSB", (checksum / 256) % 256)
```
rohbotics: Same octal comment here

rohbotics (changes requested):
Couple more things I noticed.

Also someone needs to re-test everything on a real robot.

dorkamotorka:
Applied mentioned changes. @GreTimotej do you have any cycles to test that?

GreTimotej:
Today I tested the made changes and I think everything works as it should.

rohbotics (approved):

rohbotics:
Fixed the conflicts, now merging this.

PR #135, opened 2021-06-02
Move internal docs into docs folder, add ADR template

rohbotics:
This brings ubiquity_motor more in line with the documentation standards that we are working on.


dorkamotorka:
I think the layout of the doc is great. I think we are missing a Section like "Ultimate goal" or something along the lines "What are we trying to achieve". This way the reader would know why in the first place are we even doing this.

mjstn (approved):
Nice bit of organization seems good.

PR #136, opened 2021-06-20
Merge in to 4WD from main magni code the feature that will always set zero s…

mjstn:
…peed when speed control is set to disable on topic /system_control

This merge is to address  avalon branch issue #110 
https://github.com/UbiquityRobotics/avalon/issues/110


JanezCim (approved):
Just went through code and I have no thing to complain about. Have not tested it on the real robot

mjstn:
As a reviewer it is not your task to execute the code unless it satisfies you in some way.  
In this case I did unit tests of the thing being fixed and because the way we do the stop of full control changed to be cleaner in magni I also put that in and manually tested the enable and disable there as well.    I did not test beyond raw messages to be sure the disable and enable we use in fw upgrade would work.  I did not test firmware upgrade as the thing effected was unit tested at lower level.

PR #140, opened 2021-07-19
Hack to vendor serial because it is still unreleased on noetic

rohbotics:


rohbotics:
Closing, manually merged into noetic-devel.

PR #141, opened 2021-07-20
Odom4wd scale factor correction and do wheel nulling more often

mjstn:
At this time of this PR creation the code to do the odom correction is basically using a scale factor of 1.0 so that it does not really do anything as the code is not operational yet.  

This PR is to allow a team member to take a peek at what is happening as another set of eyes.

The 2nd part of this is I am using a variable set to 0.1 rad/sec instead of a much tighter old value of 0.01 rad/sec for when the cmd_vel  Z angular rotation is a rotate of more than this we rotate but if the pure rotation z rotation drops below the value of ROTATE_TORQUE_THRESHOLD  the robot is allowed to relieve wheel stress to avoid hot wheels.


mjstn:
Made a mistake, this pull request is against 4wdChassisBranch  so I'm closing it and will form correct one next.

PR #142, opened 2021-07-20
Odom4wd scale factor for 4WD closer to reality odom.  Decouple 4wd-2wd from thin wheel option

mjstn:
Add a correction only for 4wd robot drive system that applies an odom correction for when the robot is doing a pure rotation in place.

We are also at this time decoupling the option for thin wheels from the option for use of 4wd vs 2wd.   We had been forcing the internal option of 4wd whenever thin wheels were selected.   They will be mostly separate now except for one case of when we do a default for unspecified option.


mjstn:
This is not done yet and is here for a sort of preview or second set of eyes to take a look at this correction we are forced to implement to better indicate in odom rotation the real rotation we have done due to a lot of wheel slipping.

MoffKalast (commented):
Neat, looks ok so far. Why not enable g_odom4wdRotationScale by default though?

mjstn:
This is now appearing to be functional and this PR is suitable for 1st review.   Keep in mind it works on making the assumption that the wheel slip on all wheels is symmetrical which it often will not be the case.   This hack is all we have so I feel it will make our rotational ODOM way closer to reality than before.

mjstn:
Yes, g_odom4wdRotationScale  will be set to 1.65 and enabled by default for 4WD mode.  There is a dual-purpose thing I want to clean up but since other things do it already and we want this now I will not roll it into this pull request.
You see I can check if we have thin wheels and then assume we are 4WD which I feel is a bit ugly because having thin wheels SHOULD simply modify encoder tics to odom calculations and also pid loop settings.   There really should be in motor_hardware also the option that is in motor_node.cc where we have the  2wd or 4wd but it has never made it's way to motor_hardware.cc yet.    So that will be the change I make then ask for tests once you guys get wheels.   But if it looks good enough we may just push it to Avalon to test once they are stable and I have tested all I can test in terms of time.

mjstn:
Really BIG push today to this branch.  We are trying to do these 3 things:
- Increase threshold for how we do wheel overheat stress release AND make a constant to make it easier to change later to be a ros param if that becomes required.
-  Biggest change is the odom slip scale factor for ODOM is looking good now in this code today.
- Lastly I took the plunge to make the motor_hardware layer know we were drive type of 4 wheel drive or not.  I also made v42 latest beta support setting of drive_type.  This is one of those really big changes that just had to be done to de-couple thin wheel option from 2wd or 4wd.   We have to do this even though it is risky it is really important because we just sold thin wheels to a magni owner so I want a way to support him and others besides just Avalon.  Anyway,    'petal to the metal',  and  'let the games begin'.
-

mjstn:
This is now all ready for review and pushed into branch odom4wdScaleFactor

mjstn:
Have added an informational elog for high wheel speed request so we can gain more insight into this high speed rotatation issue.  We need to determine if the motor node is being ASKED to rotate at high speed or the motor node OR mcb has 'lost its wonkers'.

src/motor_hardware.cc
```diff
@@ -50,6 +50,7 @@ const static uint8_t  I2C_PCF8574_8BIT_ADDR = 0x40; // I2C addresses are 7 bits
 // 17.2328767123 and  gear ratio of 4.29411764706:1
 #define TICKS_PER_RADIAN_ENC_3_STATE (20.50251516)    // used to read more misleading value of (41.0058030317/2)
 #define QTICKS_PER_RADIAN   (ticks_per_radian*4)      // Quadrature ticks makes code more readable later
+#define HIGH_SPEED_RADIANS  (2.0)                     // This is a threshold were we consider we are turning the wheel very fast'
```
MoffKalast: Higher than the 1.8 in the other PR.

src/motor_hardware.cc
```diff
@@ -62,6 +63,16 @@ int32_t  g_odomLeft  = 0;
 int32_t  g_odomRight = 0;
 int32_t  g_odomEvent = 0;
 
+// A 4wd robot must skid to turn. This factor approximates the actual rotation done vs what
+// the wheel encoders have indicated.  This only applies if in 4WD mode
+double   g_odom4wdRotationScale = 1.65;
+#define  WHEEL_VELOCITY_NEAR_ZERO   ((double)(0.05))
```
MoffKalast: Defined as 0.08 in the other PR.

src/motor_node.cc
```diff
@@ -486,12 +516,17 @@ int main(int argc, char* argv[]) {
...
 
             // a periodic refresh of wheel type which is a safety net due to it's importance.
-            // This can be removed when a solid message protocol is developed
             if (robot->firmware_version >= MIN_FW_WHEEL_TYPE_THIN) {
                 // Refresh the wheel type setting
                 robot->setWheelType(robot->wheel_type);
                 mcbStatusPeriodSec.sleep();
             }
+            // a periodic refresh of drive type which is a safety net due to it's importance.
+            if (robot->firmware_version >= MIN_FW_DRIVE_TYPE) {
```
MoffKalast, line 526: I take it the firmware part of this can switch at any time without major effort and sending this at a high frequency won't cause any issues?

MoffKalast (approved):
Looks good to me, I've pointed out some values which aren't the same across the two PRs, and I'd think the other one has the better ones given that we're sending it off first, so it would make sense to adjust them to the same values here too.

And I didn't see any spot where 4wd is actually defined in a launch param or such, so in case that's missing we'll need to add it.

mjstn:
Today I updated this PR to back out the warning message for high rotational rates and to back out the change of the wheel slip nulling conversion from fixed value to now a define.   In this way this PR now covers one main change, not 3.

MoffKalast (approved):
Okay good, I think we can merge if there's nothing else to add.

src/motor_hardware.cc
mjstn: I have decided upon 1.8 here.   This is not really a value that effects any performance, just limits log spam.  After some tests I decided 2 was too high so will stay with 1.8  and bingo they hit it so 1.8 is a good value.

src/motor_hardware.cc
mjstn: To error on the side of caution I'll use 0.08
This parameter will become a ros param as soon as I can get an assist.   This will be a separate PR to make this a param along with another value so somebody can do that as a separate PR.  This will be later, not right now.

src/motor_node.cc
mjstn: This code along with other refresh next to it is gated to a lower rate.  See this block as it is using sysMaintPeriod.
This is non-ideal but I do this because I had seen it and could not figure it out so this is a 'just in case' workaround.
Non-ideal but too much to try to fix in this PR.

PR #144, opened 2021-07-23
Check for mininum hardware version before reading option switch

### LINKED ISSUE #139: Wheels not moving on kinetic-devel branch
Hi folks!

I'm trying to use the kinetic-devel branch instead of the indigo-devel that is distributed with apt for kinetic for my Magni.
It works fine with indigo-devel, but when I switch to the new branch I get this error:
```
[ERROR] [1625944236.182432997]: Failed to read from I2C device /dev/i2c-1!  ERROR: Remote I/O error
[ERROR] [1625944236.183686735]: Error -2 in reading MCB option switch at 8bit Addr 0x40
```
The wheels don't spin when I publish to /cmd_vel and 
`rostopic echo /motor_power_active` gives me `data: False`

I thought it could be something with the new `system_control` topic, but publishing `motor_control enable`  and `speed_control enable ` did not seem to help.

I have this firmware version at the moment:
```
fw revision  40
fw daycode   20201209
hw revision  4.9
```
Any idéas about why this is happening? Maybe I have old hardware that is not compatible? Do I need to do something to enable the motors when using the new branch?

rohbotics:
Fixes #139 

@samiamlabs this is the initial easy part of the fix that we discussed.

### DIFF
--- a/include/ubiquity_motor/motor_hardware.h
+++ b/include/ubiquity_motor/motor_hardware.h
@@ -50,6 +50,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <gtest/gtest_prod.h>
 
+// Mininum hardware versions required for various features
+#define MIN_HW_OPTION_SWITCH 50
+
 struct MotorDiagnostics {
     MotorDiagnostics()
         : odom_update_status(
--- a/src/motor_hardware.cc
+++ b/src/motor_hardware.cc
@@ -480,6 +480,7 @@ void MotorHardware::requestSystemEvents() {
 // The 1st firmware version this is set for is 32, before it was always 1
 void MotorHardware::setHardwareVersion(int32_t hardware_version) {
     ROS_INFO("setting hardware_version to %x", (int)hardware_version);
+    this->hardware_version = hardware_version;
     MotorMessage mm;
     mm.setRegister(MotorMessage::REG_HARDWARE_VERSION);
     mm.setType(MotorMessage::TYPE_WRITE);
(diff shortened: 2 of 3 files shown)


include/ubiquity_motor/motor_hardware.h
```diff
@@ -50,6 +50,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <gtest/gtest_prod.h>
 
+// Mininum hardware versions required for various features
+#define MIN_HW_OPTION_SWITCH 50
```
mjstn, line 54: For historical reasons and visibility all firmware abilities are given around line 45 of  motor_message.h  but this is interesting as it is a function of hardware.   I think it would be good to  be here.   Most of the decisions if hardware is present or not are in the firmware itself so this is a bit of an exception.  So leave it here I think.  Just thinking outloud.

samiamlabs:
It looks to me like `robot->hardware_version` is set to 0 here and is not read from params anywhere at the moment. https://github.com/UbiquityRobotics/ubiquity_motor/blob/c4f8bbecdc66fbe06f2d34bca7fc4539a09821bd/src/motor_hardware.cc#L139
Is that correct?

include/ubiquity_motor/motor_hardware.h
rohbotics: That's the same thought process I went through :)

rohbotics:
@samiamlabs Good catch, thanks! Updated so that `setHardwareVersion` actually sets that variable. setHardwareVersion is called with the value of the ros param, and is set to 49 by default.

rohbotics:
Merging based on discussion with @mjstn at huddle yesterday.

PR #145, opened 2021-07-27
Install the shared lib for our custom serial

rohbotics:


rohbotics:
Self merging because its a fairly simple change.

PR #146, opened 2021-08-01
Use higher 4wd wheel slip threshold and warn of high rotational commands

mjstn:
Two part change:
1) Simple change of the threshold we use to decide to null excess wheel torque and thus minimize wheel overheating.
The logic is unchanged but the constant of WHEEL_SLIP_THRESHOLD.
We will later get this into the form of a motor node ros param so it can be changed.

2) A warning message to debug a violent rotational mode we are seeing after startup.

### DIFF
--- a/src/motor_hardware.cc
+++ b/src/motor_hardware.cc
@@ -50,7 +50,7 @@ const static uint8_t  I2C_PCF8574_8BIT_ADDR = 0x40; // I2C addresses are 7 bits 
 // 17.2328767123 and  gear ratio of 4.29411764706:1
 #define TICKS_PER_RADIAN_ENC_3_STATE (20.50251516)    // used to read more misleading value of (41.0058030317/2)
 #define QTICKS_PER_RADIAN   (ticks_per_radian*4)      // Quadrature ticks makes code more readable later
-#define HIGH_SPEED_RADIANS (2.0) // This is a threshold were we consider we are turning the wheel very fast'
+#define HIGH_SPEED_RADIANS (1.8) // This is a threshold were we consider we are turning the wheel very fast'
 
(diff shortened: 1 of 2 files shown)


mjstn:
Only one reviewer is required.  I put both on as it is weekend.  This is a REQUIRED change for next MoonStar update

src/motor_hardware.cc
```diff
@@ -50,6 +50,7 @@ const static uint8_t  I2C_PCF8574_8BIT_ADDR = 0x40; // I2C addresses are 7 bits
 // 17.2328767123 and  gear ratio of 4.29411764706:1
 #define TICKS_PER_RADIAN_ENC_3_STATE (20.50251516)    // used to read more misleading value of (41.0058030317/2)
 #define QTICKS_PER_RADIAN   (ticks_per_radian*4)      // Quadrature ticks makes code more readable later
+#define HIGH_SPEED_RADIANS (1.8) // This is a threshold were we consider we are turning the wheel very fast'
```
MoffKalast, line 53: Hmm that seems a bit high, 103 degrees per second? But if that's for a specific wheel, not the robot on the yaw axis I guess, so it's not that much in that respect. I assume this was tested and adjusted.

MoffKalast (approved):
Seems good, the x8 increase should hopefully help with the heating. We should also adjust move basic params for faster turning right?

src/motor_hardware.cc
mjstn: I had it lower but then rotation with the joystick can do over 1.6 so I put it here.    All this does is pepper the log with messages saying it is happening so I wanted to use a number that seemed high but it is a sort of guess.  I feel they are rotating at some rediculous value because the motor node is crying out that pwm limits are hit and so on.  So we seek notification of really high requests for rotation at high speed.

PR #148, opened 2021-08-05
[Diagnostics] Motor Power being on is not an error

rohbotics:
The error diagnostic status should be used to indicate that something is wrong,
the motor power being on is not an error, it is an expected part of
normal operation.


MoffKalast (approved):
OK.

Issue #149, opened 2021-08-09
Upgrade firmware needs to be ported to Python3


rohbotics:

JanezCim:
One shady workaround, install python2 alongside python3 and update with that

```
sudo apt install python2
curl https://bootstrap.pypa.io/pip/2.7/get-pip.py --output get-pip.py
sudo python2 get-pip.py
pip2 install requests
pip2 install pyserial
python2 ~/catkin_ws/src/ubiquity_motor/scripts/upgrade_firmware.py --file  /home/ubuntu/catkin_ws/src/ubiquity_motor/firmware/v40_20201209_enc.cyacd
```

JanezCim:
Just for future reference: Some work has been done on this on https://github.com/UbiquityRobotics/ubiquity_motor/pull/152 but was stopped. As far as i remember @rohbotics has said that this would be a bigger project including huge changes in `upgrade_firmware.py` since changes from py2 to py3 were big.

PR #152, opened 2021-09-15
Firmware update script py3 conversion

JanezCim:
Trying to solve #149 

I added `.encode()` because i was experiencing: 
```
raise TypeError('unicode strings are not supported, please encode to bytes: {!r}'.format(seq))
TypeError: unicode strings are not supported, please encode to bytes: "b'
``` 
But then I added Int() because:
```
  File "/home/ubuntu/catkin_ws/src/ubiquity_motor/scripts/upgrade_firmware.py", line 233, in send
    chr((len(self.out) / 1) % 256),
TypeError: integer argument expected, got float
``` 

And now im getting:
```
  File "/home/ubuntu/catkin_ws/src/ubiquity_motor/scripts/upgrade_firmware.py", line 261, in send
    raise Exception("Packet did not start with valid start byte, instead: " + cstr(start_byte) + " -- header: " + cstr(header))
Exception: Packet did not start with valid start byte, instead: <class 'bytes'> -> b'~' -- header: <class 'bytes'> -> b'~<0\x00'
 
```

and I cant solve it. Do you have any ideas how to go around this @rohbotics? I maybe going at it from a totally wrong angle?

### DIFF
--- a/scripts/upgrade_firmware.py
+++ b/scripts/upgrade_firmware.py
@@ -82,7 +82,7 @@ print("\nUpdating firmware now. Do not power off the robot. This is expected to 
 
 
 # Begin the code firmware uploading code
-DEBUG = False
+DEBUG = True
 class InvalidFileException(Exception):
     def __init__(self):
         pass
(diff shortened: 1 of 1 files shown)


mjstn:
Rohan has been converting the firmware upgrade script for a few weeks now.  This is a duplicate issue of some other issue.

JanezCim:
yes you mean #149 , but this is a draft PR, not an issue, here im trying to colaborate with rohan to actually solve it. And as far as I could see there was no branch to this goal. I got myself into this because im working on OTA update for Noetic and this script first needs to work in order for that to work.

JanezCim:
just FYI on ota_update this has been kindof patched here https://github.com/UbiquityRobotics/ota_update/pull/13 by simply installing python2 onto the same system. This is a very bad solution long term, but for now it works. When Firmware update script gets updated to py3, this should immediately be fixed also on ota_update repo

JanezCim:
This py2 hack is now also planned on being used in firmware images in https://github.com/UbiquityRobotics/pi_image2 so when py3 conversion is done, someone should also check to update the commands there

PR #153, opened 2021-09-20
Merge4wd and gear ratio to kinetic devel

mjstn:
Although this code has been merged to noetic-devel it seems a good plan to also put the changes from merge4wdFeaturesToMagni along with addMotorGearRatioParameter in this PR.
The reason it is being done as a separate PR is due to things in kinetic-devel that vary from noetic-devel and this one PR makes the total set of changes clear from the perspective of the shipping product code of today.


mjstn:
I put you on here Vid because Janez did last one and I'm  spreading the joy.  Also something showed up related to tables for battery so you may want to verify I did not mess that up.

mjstn:
Removing this PR now because a flawed merge operation that formed it is not usable in kinetic-devel.

PR #155, opened 2021-10-12
Update4wd and gear ratio

mjstn:
This updates the standard Magni codebase with many improvements and options not present in the past as follows:

- Publish a motor_state topic to show low level motor currents and individual wheel positions and rotate rates and pwm drive
- Adds a  pid_control setting to enable special modes in the firmware mostly of value for 4 wheel drive modes
- Adds a 'drive_type'  to yaml config file that defaults to 'standard' (also the current firmware_default value). Can be 4wd
- Adds a 'wheel_type' to yaml config which defaults to 'standard' but for some custom hardware can be set to 'thin'
- Adds a wheel_gear_ratio' to yaml config that is normally 4.294.  We may use other wheels where this is 5.17
- Adds a 'fw_max_pwm' to yaml that hard sets max PWM value set in firmware as a cap for illegal states

Other assorted improvements to log status info and some other things come in with this 'large' update.


mjstn:
A few minor things have been pulled into this due to it taking a while to get reviewed.  Nothing all that big.
I have added comments around the cm.update call because this HUGE mystery is totally 'glossed over' and EXTREMELY complex so a nice fat comment is rather important.

README.md
```diff
@@ -90,6 +99,21 @@ Divisor for the above `PID` paramaters.
 `pid_moving_buffer_size` (int, default: 10)
 Size of a moving buffer used in the control loop.
 
+`pid_control` (int, default: 0)
+This word enables non-standard modes for motor control in the firmware. Consult factor for guidance in usage of this parameter.
```
MoffKalast: consult factor?

include/ubiquity_motor/motor_parameters.h
```diff
@@ -185,15 +194,25 @@ struct NodeParams {
...
 
     int mcbControlEnabled;    // State to allow suspension of MCB serial control for diagnostic purposes
     int mcbSpeedEnabled;      // State to allow zero speed override for safety reasons
 
-    NodeParams() : controller_loop_rate(10.0), wheel_type("firmware_default"), wheel_direction("firmware_default"),
-        mcbControlEnabled(1), mcbSpeedEnabled(1){};
+    NodeParams() : controller_loop_rate(10.0),
+        wheel_type("firmware_default"), 
+        wheel_direction("firmware_default"),
+	wheel_gear_ratio(WHEEL_GEAR_RATIO_1),
```
MoffKalast: Mixed indents of tabs and spaces in different lines, may be worthwhile to run a convert to one type at some point for better readability.

src/motor_node.cc
```diff
@@ -47,6 +47,9 @@ static FirmwareParams g_firmware_params;
 static CommsParams    g_serial_params;
 static NodeParams     g_node_params;
 
+// TODO: Enhancement - Make WHEEL_SLIP_THRESHOLD be a ROS param
+#define WHEEL_SLIP_THRESHOLD  (0.08)   // Rotation below which we null excess wheel torque in 4wd drive_type
```
MoffKalast, line 51: Wouldn't this be more like motor stall threshold? Also affects heavily loaded 2wd robots and could prevent the occasional annoying buzz I'd imagine?

src/motor_node.cc
```diff
@@ -143,6 +146,12 @@ void initMcbParameters(std::unique_ptr<MotorHardware> &robot )
         } else if (g_node_params.wheel_type == "thin"){
             wheel_type = MotorMessage::OPT_WHEEL_TYPE_THIN;
             ROS_INFO("Host is specifying wheel_type of '%s'", "thin");
+
+            // If thin wheels and no drive_type is in yaml file we will use 4wd
+            if (g_node_params.drive_type == "firmware_default") {
```
MoffKalast, line 151: The better option would be to go into the 4wd repo and make sure the drive type is defined as 4wd there, but assuming nobody else is using thin wheels right now I guess we can leave it at that.

src/motor_node.cc
```diff
@@ -189,9 +230,19 @@ void initMcbParameters(std::unique_ptr<MotorHardware> &robot )
...
+    // For such robots when loaded down the wheels can get in a state where they cannot skid.
+    // This leads to motor overheating.  This code below sacrifices accurate odometry which
+    // is not achievable in such robots anyway to relieve high wattage drive power when zero velocity.
+    g_wheel_slip_nulling = 0;
+    if ((robot->firmware_version >= MIN_FW_WHEEL_NULL_ERROR) && (g_node_params.drive_type == "4wd")) {
+        g_wheel_slip_nulling = 1;
+        ROS_INFO("Wheel slip nulling will be enabled for this 4wd system when velocity remains at zero.");
+    }
+
     // Tell the MCB board what the I2C port on it is set to (mcb cannot read it's own switchs!)
```
MoffKalast: Wait, the MCB doesn't know it's i2c address? What? Do we... use i2c on the MCB?

src/motor_node.cc
```diff
@@ -341,11 +401,13 @@ int main(int argc, char* argv[]) {
     ros::Time last_sys_maint_time = last_loop_time;
     ros::Time last_joint_time = last_loop_time;
     ctrlLoopDelay.sleep();                  // Do delay to setup periodic loop delays
+    uint32_t loopIdx = 0;
 
     while (ros::ok()) {
         current_time = ros::Time::now();
         elapsed_loop_time = current_time - last_loop_time;
         last_loop_time = current_time;
+	loopIdx += 1;
```
MoffKalast: If the loop runs for 497 days at 50 Hz this would technically overflow I think. Not sure if that's handled or if it would even cause any issues conceptually. So it may not be a thing worth thinking about at all since 1000 things would likely go wrong until then I guess, but some people may be running their robots 24/7 in some applications so it could conceivably be reached.

src/motor_node.cc
```diff
@@ -443,6 +546,12 @@ int main(int argc, char* argv[]) {
                 robot->setWheelType(robot->wheel_type);
                 mcbStatusPeriodSec.sleep();
             }
+            // a periodic refresh of drive type which is a safety net due to it's importance.
```
MoffKalast, line 549: Just in case the robot suddenly grows two extra wheels 😛

MoffKalast (approved):
Seems good to merge, mostly just nitpicks.

README.md
mjstn: Fixed: 'factory'

include/ubiquity_motor/motor_parameters.h
mjstn: I do really HATE tabstops.  They are an evil thing invented when computers had 200megabyte main disks.  Thanks

src/motor_node.cc
mjstn: This is sort of a just in case for something I had seen but cannot figure out where this was not set and all heck broke loose.

src/motor_node.cc
mjstn: Yes but note it is 'unsigned' and also the index is only used in a modulo way so 'all is well'.

src/motor_node.cc
mjstn: The MCB was designed with a puny little micro so it is way way out of pins.  Thus this is UGLY as all get-out but sadly is required.   Host reads the option port on I2C and then tells a register in the MCB what it's board rev is.  Don't get me started.   On all Gen6 boards I have insisted with a heavy hand that ALL boards with processors will read and know their own hardware revision and THEN report THAT to the host.

src/motor_node.cc
```diff
@@ -240,7 +240,7 @@ void initMcbParameters(std::unique_ptr<MotorHardware> &robot )
         ROS_INFO("Wheel slip nulling will be enabled for this 4wd system when velocity remains at zero.");
     }
 
-    // Tell the MCB board what the I2C port on it is set to (mcb cannot read it's own switchs!)
+    // Tell the MCB board what the port that is on the Pi I2c says on it (the mcb cannot read it's own switchs!)
```
MoffKalast, line 243: I still don't quite get what this is about exactly, but I think we can leave it for some other time.

MoffKalast (approved):

src/motor_node.cc
MoffKalast: Ah I see, well I sure hope we go for some spare flash memory and more pins next time 😄

PR #168, opened 2022-03-17
Noetic Battery Percentage, Firmware Topic Pub

MoffKalast:
This PR imports the changes from PR https://github.com/UbiquityRobotics/ubiquity_motor/pull/138 which was only merged with kinetic-devel (though some fragments somehow made it into noetic-devel somehow). The voltage levels have been updated to reflect a customer battery characterization study.

Also to provide easier access to the firmware version and date I've set up a String publisher for it. Required by https://github.com/UbiquityRobotics/oled_display_node/pull/22

### DIFF
--- a/src/motor_hardware.cc
+++ b/src/motor_hardware.cc
@@ -30,7 +30,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ubiquity_motor/motor_hardware.h>
 #include <ubiquity_motor/motor_message.h>
 #include <boost/assign.hpp>
-
 #include <boost/math/special_functions/round.hpp>
 
 // To access I2C we need some system includes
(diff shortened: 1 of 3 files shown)


src/motor_hardware.cc
```diff
@@ -285,15 +286,18 @@ void MotorHardware::readInputs(uint32_t index) {
                     } else {
                         ROS_INFO_ONCE("Firmware version %d", mm.getData());
                         firmware_version = mm.getData();
-			motor_diag_.firmware_version = firmware_version;
+                        motor_diag_.firmware_version = firmware_version;
                     }
+                    publishFirmwareInfo();
```
mjstn, line 291: This only posts one time but oled does get the info.  LEAVE it for now and at some later time I will do the part that requires export of motor_hardware api over to motor_node.cc for periodic firmware publish.   Don't mess with it, lets just push this puppy.

mjstn (approved):
This looks really nice Vid.   A big improvement.   Lets ship it!

src/motor_hardware.cc
MoffKalast: Yep, I've had it publish every message at first but it was just way too much spam. So I made it latched instead and every one will get the message once it subscribes to the topic. It does actually seem to get published 3 times (first twice with only the version and then once more with the date which then gets latched).

https://github.com/UbiquityRobotics/ubiquity_motor/blob/5b542a55fd133307150c107d0879a8c42909a598/src/motor_hardware.cc#L152

PR #173, opened 2023-10-09
Implemented power supply status estimation

ahrastnik:
This adds the ability to estimate the charging status, which is necessary information for autonomous charging.

Implementation details along with simulation, research data / measurements and limitations can be found [here](https://github.com/UbiquityRobotics/charge_status_research).

### DIFF
--- a/src/motor_hardware.cc
+++ b/src/motor_hardware.cc
@@ -31,6 +31,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ubiquity_motor/motor_message.h>
 #include <boost/assign.hpp>
 #include <boost/math/special_functions/round.hpp>
+#include <boost/circular_buffer.hpp>
 
 // To access I2C we need some system includes
 #include <linux/i2c-dev.h>
@@ -68,6 +69,8 @@ int32_t  g_odomLeft  = 0;
 int32_t  g_odomRight = 0;
 int32_t  g_odomEvent = 0;
 
+// 2x6S(12 V) Batteries connected in series
+const uint8_t SLA_AGM_CELLS = 6 * 2;
 //lead acid battery percentage levels for a single cell
 const static float SLA_AGM[11] = {
     1.800, // 0
@@ -83,6 +86,9 @@ const static float SLA_AGM[11] = {
     2.175, // 100
 };
 
+const float SLA_AGM_V_NOM = 2.175f * SLA_AGM_CELLS; // NOMINAL battery bank voltage
+const float SLA_AGM_V_CHG = 2.45f * SLA_AGM_CELLS; // CHARGING battery bank voltage
+const float SLA_AGM_V_TRI = 2.333f * SLA_AGM_CELLS; // Trickle charging battery bank voltage
 
 //li-ion battery percentage levels for a single cell
 const static float LI_ION[11] = {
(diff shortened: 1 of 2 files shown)


src/motor_hardware.cc
```diff
@@ -607,6 +613,93 @@ float MotorHardware::calculateBatteryPercentage(float voltage, int cells, const
...
+            status = sensor_msgs::BatteryState::POWER_SUPPLY_STATUS_FULL;
+        } else {
+            trickle_dur += dt_r;
+        }
+    } else if (is_v_dropping && is_in_nom_range && !is_charging) {
+        // Discharging
+        status = sensor_msgs::BatteryState::POWER_SUPPLY_STATUS_DISCHARGING;
+        trickle_dur = ros::Duration(0);
+    }
+    printf("%f, %f, %f, %f, %f, %d\n", ros::Time::now().toSec(), battery_voltage, voltage_avg, d, trickle_dur.toSec(), status);
```
mjstn: We don't use raw printf except prior to start of ROS situations.  See other uses in this same file of  ROS_INFO

ahrastnik: Oh shoot, I forgot to remove that after debugging - will remove.

src/motor_hardware.cc
```diff
@@ -68,6 +69,8 @@ int32_t  g_odomLeft  = 0;
 int32_t  g_odomRight = 0;
 int32_t  g_odomEvent = 0;
 
+// 2x6S(12 V) Batteries connected in series
+const uint8_t SLA_AGM_CELLS = 6 * 2;
```
mjstn, line 73: I suppose you could use this but any of our batteries alone at 12V is 'low'.   I do see later you use 2.175 sort of number that would be 13V.  A nominal 12V gel cell is near 12.6V actually.
I leave this up to you to think about and you can leave this if you like except the 2.175 factor should be a factor to come out to more like 12.6 but again batteries brand new fresh totally charted can read 13v so up to you.

src/motor_hardware.cc
ahrastnik: Removed.

mjstn:
I would say this is ok as long as you convert the printf to be a ROS_INFO call because if you don't we don't see this message in the log.    Now maybe that printf was debug code, up to you.  If it is debug code then use of ROS_DEBUG would be used if you want that debug code to be seen in the future.  keep in mind that if you hammer this routine all the time ROS_INFO is bad and so is printf because it will 'spam like crazy'.   In the past I have used other more tricky things like have some static counter and only do the ROS_INFO every 5 seconds worth of time or something like that.

src/motor_hardware.cc
ahrastnik: If you are referring to `SLA_AGM_CELLS`, this number only indicates the number of cells in the used battery bank - and is not used directly anywhere in the code, but rather only to determine voltages levels with factors as seen below.

Perhaps my choice in variable names wasn't great, but these three constants below are used only to determine the charging state and I found that these factors work best for that purpose alone.

ahrastnik:
I don't think that is necessary. This code gets executed at about 7 Hz and
it will spam and fill the logs if viewed at a debug level. Although,
perhaps it would be wise to output only when the charging state is changed
- I don't have the time to implement this, unfortunately, but it's easy
enough to do for whoever picks up the code next.

V V sre., 11. okt. 2023 ob 10:15 je oseba Mark Johnston <
***@***.***> napisala:

